PR #5, opened 2015-04-20
Add support for bug-reports URL

sol:


src/Config.hs
```diff
@@ -50,6 +50,7 @@ data ConfigFile = ConfigFile {
 , configFileVersion :: Maybe String
 , configFileSynopsis :: Maybe String
 , configFileDescription :: Maybe String
+, configFileBugReports :: Maybe String
```
soenkehahn: Could we rename `ConfigFile` in `PackageDescription` and `readConfig` in something appropriate?

soenkehahn:
LGTM.

src/Config.hs
sol: Rename it to `PackageConfig` and `readPackageConfig`.

PR #7, opened 2015-04-23
detect .lhs files for other-modules

soenkehahn:


sol:
Looks good, thanks a lot!

PR #12, opened 2015-05-01
Warn on source-dirs which do not exist

### LINKED ISSUE #11: Confusing source-dirs
Some `package.yaml` file use e.g. `driver, src` as the value for `source-dirs`. This is not a valid yaml list value.

It still works most of the time because the field value gets inserted without quotes into the cabal file. For example: the `package.yaml` of this repo.

Should `hpack` warn about non-existing source directories?

Fuuzetsu:
Closes #11

### DIFF
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -1,4 +1,4 @@
--- This file has been generated from package.yaml by hpack version 0.1.1.
+-- This file has been generated from package.yaml by hpack version 0.1.2.
 --
 -- see: https://github.com/sol/hpack
 
@@ -57,5 +57,6 @@ test-suite spec
     , hspec == 2.*
     , mockery
     , interpolate
+    , silently
   ghc-options: -Wall
   default-language: Haskell2010
--- a/package.yaml
+++ b/package.yaml
@@ -32,3 +32,4 @@ tests:
       - hspec == 2.*
       - mockery
       - interpolate
+      - silently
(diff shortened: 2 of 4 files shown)


jkarni:
Maybe a test case?

Fuuzetsu:
I would write one but I don't know of a good way to test stdout. I'll look into it though.

jkarni:
[silently](https://hackage.haskell.org/package/silently-1.2.4.1) might do it

Fuuzetsu:
Nice package, see the most recent commit in this PR.

I do wonder whether we need some kind of verbosity setting however. I also wonder whether a `Writer` wouldn't do a nicer job though I don't think that's a concern unless `hpack` is exposed as a library too.

sol:
> I also wonder whether a Writer wouldn't do a nicer job though I don't think that's a concern unless hpack is exposed as a library too.
> 

For me code quality is always a concern, not only for libraries.  We don't necessarily need a Writer for that, just return a list of warnings with the package specification.

In addition:

 - There is a bug, if you have a top-level source-dir and two executables (or a library and an executable), I think you get duplicate warnings.
- Do we currently get warnings for test suites?
- Warnings should go to stderr
- If your code is hard to test, there maybe something wrong with your code,  one of the advantages of TDD is that it leads to cleaner, more functional design.  Writing to stdout while parsing a package specification I consider bad design. Output is currently only produced in Main.hs, and I would prefer to keep it that way.

sol:
Closing in favor of #15.

PR #13, opened 2015-05-05
Warn on unknown fields in package description

sol:


src/Config.hs
```diff
@@ -19,11 +21,28 @@ import           Data.Yaml
...
 import           Data.HashMap.Lazy (HashMap)
 import qualified Data.HashMap.Lazy as Map
+import qualified Data.Text as T
 import           System.FilePath
 import           System.Directory
+import           Data.Data
 
 import           Util
 
+data CaptureUnknownFields a = CaptureUnknownFields [String] a
```
soenkehahn: This is nice!

src/Util.hs
```diff
@@ -21,15 +23,19 @@ import           Data.Maybe
...
 
 import           GHC.Generics
 import           Data.Aeson.Types
 
 genericParseJSON_ :: (Generic a, GFromJSON (Rep a)) => String -> Value -> Parser a
-genericParseJSON_ name = genericParseJSON defaultOptions {fieldLabelModifier = camelTo '-' . drop (length name)}
+genericParseJSON_ name = genericParseJSON defaultOptions {fieldLabelModifier = hyphenize name}
+
+hyphenize :: String -> String -> String
+hyphenize name = camelTo '-' . drop (length name)
```
soenkehahn: I think `hyphenize` sounds too general. Maybe something like `toJSONField`?

soenkehahn:
Apart from comments LGTM.

PR #14, opened 2015-05-08
Warn on unknown fields in library/executable/test sections

sol:


src/Config.hs
```diff
@@ -25,10 +27,20 @@ import qualified Data.Text as T
...
-data CaptureUnknownFields a = CaptureUnknownFields [String] a
+genericParseJSON_ :: forall a. (Typeable a, Generic a, GFromJSON (Rep a)) => Value -> Parser a
+genericParseJSON_ = genericParseJSON defaultOptions {fieldLabelModifier = hyphenize name}
+  where
+    name = (tyConName . typeRepTyCon . typeRep) (Proxy :: Proxy a)
+
+hyphenize :: String -> String -> String
+hyphenize name = camelTo '-' . drop (length name)
+
+data CaptureUnknownFields a = CaptureUnknownFields {captureUnknownFieldsFields :: [String], captureUnknownFieldsValue :: a}
```
soenkehahn: Could we break up this line?

src/Config.hs
```diff
@@ -170,9 +182,9 @@ mkPackage (CaptureUnknownFields unknownFields PackageConfig{..}) = do
   let sourceDirs = fromMaybeList packageConfigSourceDirs
   let defaultExtensions = fromMaybeList packageConfigDefaultExtensions
   let ghcOptions = fromMaybeList packageConfigGhcOptions
-  mLibrary <- mapM (mkLibrary sourceDirs dependencies defaultExtensions ghcOptions) packageConfigLibrary
-  executables <- toExecutables sourceDirs dependencies defaultExtensions ghcOptions packageConfigExecutables
-  tests <- toExecutables sourceDirs dependencies defaultExtensions ghcOptions packageConfigTests
+  mLibrary <- mapM (mkLibrary sourceDirs dependencies defaultExtensions ghcOptions) (captureUnknownFieldsValue <$> packageConfigLibrary)
+  executables <- toExecutables sourceDirs dependencies defaultExtensions ghcOptions (fmap captureUnknownFieldsValue <$> packageConfigExecutables)
+  tests <- toExecutables sourceDirs dependencies defaultExtensions ghcOptions (fmap captureUnknownFieldsValue <$> packageConfigTests)
```
soenkehahn: Maybe break up these lines too?

soenkehahn:
Apart from comments LGTM.

src/Config.hs
sol: Yes, good point.

src/Config.hs
sol: Done in 926e3f3f9860715a730dd1086e4ff4b7916dcce6

PR #15, opened 2015-05-14
Warn on missing source dirs (fixes #11)

### LINKED ISSUE #11: Confusing source-dirs

sol:


soenkehahn:
LGTM.

PR #16, opened 2015-05-14
Accept arbitrary entry points for executables

sol:


src/Haskell.hs
```diff
@@ -0,0 +1,28 @@
...
+isModuleName name = case name of
+  x : xs -> isUpper x && all isIdChar xs
+  _ -> False
+
+isQualifiedIdentifier :: [String] -> Bool
+isQualifiedIdentifier name = case reverse name of
+  x : xs  -> isIdentifier x && isModule xs
+  _ -> False
+
+isIdentifier :: String -> Bool
```
jkarni: Isn't this more properly a `isVariableIdentifier`? Constructor identifiers can have uppercase

sol: Good point.  I'll also reject reserved keywords.

sol: Hmm, but the it would become `isQualifiedVariableIdentifier`.  I think I will ignore the ambiguity in naming for now until we actual have a name clash.

test/HaskellSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+
+    it "rejects the empty list" $ do
+      isModule [] `shouldBe` False
+
+  describe "isQualifiedIdentifier" $ do
+    it "accepts qualified Haskell identifiers" $ do
+      isQualifiedIdentifier ["Foo", "Bar", "baz"] `shouldBe` True
+
+    it "rejects invalid input" $ do
+      isQualifiedIdentifier ["Foo", "Bar", "Baz"] `shouldBe` False
```
soenkehahn: Maybe add a test for `["Foo", "bar", "Baz"]`?

src/Util.hs
```diff
@@ -24,31 +26,42 @@ import           Data.Data
...
+
+parseMain :: String -> (FilePath, [GhcOption])
+parseMain main = case reverse name of
+  x : _ | isQualifiedIdentifier name && x `notElem` ["hs", "lhs"] -> (intercalate "/" (init name) ++ ".hs", ["-main-is " ++ main])
+  _ | isModule name -> (intercalate "/" name ++ ".hs", ["-main-is " ++ main])
+  _ -> (main, [])
+  where
+    name = splitOn '.' main
+
+splitOn :: Char -> String -> [String]
```
soenkehahn: This could have a more generic type: `a -> [a] -> [[a]]`. Don't know if that would be good.

There is `wordsBy` in the `split` package. Maybe you want to use the same name? Although, it takes a predicate as first argument.

soenkehahn:
This seems neat, but I wonder: what's the primary motivation for supporting something like `Run.run`?

I also wonder which ghc versions support this.

jkarni:
Some relevant discussion [here](https://github.com/haskell/cabal/issues/172). 

> This seems neat, but I wonder: what's the primary motivation for supporting something like Run.run?

I kind of like the idea of being allowed to have two or more different executables from only one source file.

soenkehahn:
This would also remove the necessity for having dummy `Main.hs` files, right?

soenkehahn:
LGTM.

soenkehahn:
Yeah, I want this. :)

sol:
Yes, me too. But there is still a concern, according to the GHC manual `-main-is` is a compile time option. As I understand it, objects files compiled with `-main-is` can not be reused for other purposes and object files compiled without it can not be used as main entry point  (but I still have to try). This ok if you already have a double compilation setup setup, but e.g having a library section and the depend on that in an executable and specify one module from the library as main probably won't work :(

soenkehahn:
You mean it will just break? As in cabal will through link errors?

sol:
`cabal` will complain about a missing source file.  Haven't tried plain `ghc`.

sol:
I still think it's possible to get this working by generating a `Main.hs` instead of relying on GHCs `-main-is`.

But I'm not sure if I want to do this right now.  This would be a super useful feature, but we would diverge from "simple generating cabal files" at that point.

PR #17, opened 2015-05-31
Use package.yml instead of package.yaml

sol:


sol:
@soenkehahn With this change `package.yaml` will not work anymore.  Do you think we need a deprecation cycle?

soenkehahn:
I think I would add a warning that `hpack` is still experimental to the `README` and go without deprecation cycle.

cc: @qoelet

qoelet:
LGTM

PR #20, opened 2015-06-02
Give feedback to the user running the binary

### LINKED ISSUE #18: Command line interface is not particularly user-friendly
It would be helpful if hpack had shown *anything* when run from the command line, even if it's just `Generating foo.cabal from package.yml` or a usage string. The current behavior is confusing:

```
% hpack
% hpack --help
% hpack help
%
```

supki:
An attempt to fix #18

### DIFF
--- a/driver/Main.hs
+++ b/driver/Main.hs
@@ -55,7 +55,7 @@ main = do
 
 printUsage :: [String] -> IO ()
 printUsage errors = do
-  hPutStrLn stderr (intercalate "\n" (usage : errors))
+  hPutStr stderr (concat (usage : errors))
   exitFailure
 
 printHelp :: IO ()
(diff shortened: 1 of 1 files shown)


driver/Main.hs
```diff
@@ -21,12 +24,39 @@ header = unlines [
...
+  ]
+
+usage :: String
+usage = usageInfo ("hpack " ++ showVersion version) options
+
 main :: IO ()
 main = do
+  (opts, args, errs) <- fmap (getOpt Permute options) getArgs
+  unless (null errs) (printUsage errs)
+  unless (null args) (printUsage ["hpack does not accept non-option arguments\n"])
```
soenkehahn: This results in two `\n` in the output, doesn't it?

supki: Right. There are also extra empty lines between error messages. I think it's reasonable to remove them too.

sol:
@supki Sorry for the late reply.  Can you move this into a separate module under `src`?  `driver/Main.hs` is meant to be minimal (yes I know, we already have logic in there, but we don't want to make it worse).  Some tests would also be nice ;).

PR #24, opened 2015-06-26
Glob extra source files

Fuuzetsu:


soenkehahn:
The current tests for `readPackageConfig` are already quite lengthy. As this is a new feature this could be easily tested separately. E.g. would it make sense to have a function
``` haskell
expandGlobs :: [String] -> IO ([Warnings], [FilePath])
```
that we test separately?

What about patterns like `*.js`? Do they work?

Does `foo/*.js` currently find `foo/bar/baz.js`? What about `**/*.js` for recursive file matching?

test/ConfigSpec.hs
```diff
@@ -172,6 +175,60 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+    it "warns if an extra-source-file doesn't match anything" $ do
+      writeFile "package.yaml" [i|
+        extra-source-files:
+          - missing.foo
+          - res/*
+        |]
+      fmap fst <$> readPackageConfig "package.yaml" `shouldReturn` Right [
+            "Specified extra-source-file \"missing.foo\" does not exist, leaving as-is"
+          , "Specified extra-source-file \"res/*\" does not exist, leaving as-is"
+          ]
```
soenkehahn: Do the non-matching patterns end up in the cabal file? What about dismissing them?

Fuuzetsu:
>expandGlobs :: [String] -> IO ([Warnings], [FilePath])

Sound good.

>Does foo/*.js currently find foo/bar/baz.js? What about **/*.js for recursive file matching?

No to both.

test/ConfigSpec.hs
Fuuzetsu: Yes, they do, seemed like the most logical thing to do to me and it doesn't change existing behaviour. They could be dismissed if desired though.

Fuuzetsu:
Moved out the function and testing it separately now.

Do you want the recursive file matching or were you just asking in general?

soenkehahn:
I've talked about this with @sol yesterday. IIRC our conclusion was:

1. We'd like to have recursive matching with e.g. `**/*.js`.
2. In case a pattern doesn't match any results we want to
  1. omit it from the cabal file and
  2. issues a warning.
3. Same goes for files. So if e.g. `foo.txt` is given in `extra-source-files` but cannot be found in the file system we want to omit it from the cabal file and issue a warning.

@sol: Please correct me if I'm wrong.

Fuuzetsu:
1. Done
2. Done with caveat that redundant patterns are not subject to warning, so `foo/*` and `foo/hello` don't produce a warning, provided `foo/hello` exists.
3. Exactly the same as 2, there's no distinction between what's a file path and what's a pattern

I had to switch packages to `Glob`. `filemanip` did `**` fine but that part of the package broke `*/` style patterns so we could have one or the other but not both at once.

Leaving the commits as-is, feel free to squash.

sol:
This code is much more complicated than it needs to be.  Closing in favor of #27.

Fuuzetsu:
The complexity arised from not wanting to traverse the whole directory tree for each pattern, something #27 does. Consider network shares &c.

But if that's OK then yes, code becomes simple.

sol:
Good point, yet, this code is close to incomprehensible to me.  For me the question is how many glob patterns a `package.yaml` will typically use, and how big the directory tree is in practice.

Did you observe any indications for performance issues?

If not, I would stick to Knuth's advice: "premature optimization is the root of all evil (or at least most of it) in programming"

sol:
Hmm, I admit that it feels wrong to go with the current approach considering that somebody may list many files explicitly, without using any blogs. If we can come up with code that is similar to what we currently have (in terms of complexly), but only reads the tree once, I would prefer that.

Fuuzetsu:
@sol Actually I just thought of a simple way to have both, simple code and no extra traversals, going to send a quick PR in a second

sol:
If you pass an empty list of patterns to globDir, it does a recursive directory listing. That may be useful ;)

PR #25, opened 2015-06-28
Git subdir

Fuuzetsu:


jkarni:
LGTM - the only minor quibble is that maybe a new datatype would be clearer/safer than `(String, Maybe String)`

hpack.cabal
```diff
@@ -31,6 +31,7 @@ executable hpack
     , deepseq
     , directory
     , filepath
+    , split
```
sol: I think `text` also provides the functionality we need from `split` and is already a transitive dependency, so I would prefer to use that.

test/RunSpec.hs
```diff
@@ -102,6 +102,19 @@ spec = do
...
+          "name: foo"
+        , "version: 0.0.0"
+        , "build-type: Simple"
+        , "cabal-version: >= 1.10"
+        , ""
+        , "source-repository head"
+        , "  type: git"
+        , "  location: https://github.com/hspec/hspec"
+        , "  subdir: driver"
+        ]
```
sol: Do you think we could test `renderSourceRepository` independently?  I think the current testing approach does not scale.

test/ConfigSpec.hs
```diff
@@ -177,7 +177,16 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+      packageSourceRepository c `shouldBe`
+        Just ("https://github.com/hspec/hspec", Nothing)
+
+    it "accepts third part of github URL as subdir" $ do
+      writeFile "package.yaml" [i|
+        github: hspec/hspec/driver
+        |]
+      Right (_, c) <- readPackageConfig "package.yaml"
+      packageSourceRepository c `shouldBe`
+        Just ("https://github.com/hspec/hspec", Just "driver")
```
sol: If you change `driver` to `hspec-core` the example is more realistic ;).

src/Config.hs
```diff
@@ -159,7 +162,7 @@ data Package = Package {
 , packageLicense :: Maybe String
 , packageLicenseFile :: Maybe FilePath
 , packageExtraSourceFiles :: [FilePath]
-, packageSourceRepository :: Maybe String
+, packageSourceRepository :: Maybe (String, Maybe String)
```
sol: I think at this point I'm also in favor of adding a proper `SourceRepository` type (kudos to @jkarni).

Fuuzetsu:
Addressed the comments in new commits. Once everything is good I'll squash it of course.

Regarding `splitOn`, I replaced it with `Data.Text.split` as advised. There were multiple ways to do it: 

1. parse out Text from the yaml file, split what we get out and from there just use `String` so everything else works
2. Do the above but without converting to String. This made it a bit ugly as we were now mixing String and Text here and there.
3. Parse out as `String` but convert to `Text` just for the split.

I opted for 1. Option 3 would work too and would keep the `PackageConfig` field type unchanged if that's desired, very simple change. Option 2 could be nicer if we went ahead and refactored everything to just use `Text`.

src/Config.hs
```diff
@@ -186,6 +188,16 @@ data Executable = Executable {
 , executableCppOptions :: [CppOption]
 } deriving (Eq, Show)
 
+data GithubConfig = GithubConfig {
+  githubConfigUrl :: String
+, githubConfigSubdir :: Maybe String
+} deriving (Eq, Show)
```
sol: Can we rename this type into `SourceRepository`?

src/Config.hs
```diff
@@ -261,21 +273,28 @@ mkPackage (CaptureUnknownFields unknownFields PackageConfig{..}) = do
       where
         f name = "Specified source-dir " ++ show name ++ " does not exist"
 
-    github = ("https://github.com/" ++) <$> packageConfigGithub
+    github :: Maybe GithubConfig
+    github = modifyGithubUrl ("https://github.com/" <>) <$>
+             maybe Nothing (Just . splitGithub) packageConfigGithub
```
sol: ^ please don't align

src/Config.hs
```diff
@@ -261,21 +273,28 @@ mkPackage (CaptureUnknownFields unknownFields PackageConfig{..}) = do
...
+      where
+        splitGithub str = case T.unpack <$> T.split (== '/') str of
```
sol: Using `fmap` on lists I consider code obfuscation.  See also https://ro-che.info/articles/2015-02-12-foldable-traversable-parametricity by @feuerbach.

src/Config.hs
```diff
@@ -261,21 +273,28 @@ mkPackage (CaptureUnknownFields unknownFields PackageConfig{..}) = do
...
+            GithubConfig (user <> "/" <> repo) (Just subdir)
+          _ -> GithubConfig (T.unpack str) Nothing
 
     homepage :: Maybe String
     homepage = case packageConfigHomepage of
       Just Nothing -> Nothing
       _ -> join packageConfigHomepage <|> fromGithub
       where
-        fromGithub = ((++ "#readme") <$> github)
+        fromGithub = (<> "#readme") . githubConfigUrl <$> github
```
sol: As we stick with `String` I would prefer to not change this to `<>` (as `<>` is not available in older versions of base).

test/RunSpec.hs
```diff
@@ -150,3 +138,18 @@ spec = do
           , "  ghc-options: -Wall -Werror"
           , "  default-language: Haskell2010"
           ]
+
```
sol: The following specs should be group under `describe "renderSourceRepository"`

test/RunSpec.hs
```diff
@@ -150,3 +138,18 @@ spec = do
...
+    it "renders source-repository without subdir correctly" $ do
+      renderSourceRepository (GithubConfig "https://github.com/hspec/hspec" Nothing)
+        `shouldBe` unlines [ "source-repository head"
+                           , "  type: git"
+                           , "  location: https://github.com/hspec/hspec"
+                           ]
```
sol: ^ please don't align.  Use e.g. the following instead:

```
`shouldBe` unlines [
    "source-repository head"
  , "  type: git"
  , "  location: https://github.com/hspec/hspec"
  ] 
```

sol: My reason is that if you change this code to e.g. `(T.pack . unlines) [...` later on, you have to realign the whole list.

sol:
@Fuuzetsu Yes, I think your approach regarding `String` and `Text` is good.  Other then my comments, LGTM.

Fuuzetsu:
See latest commit.

PR #28, opened 2015-07-05
Only traverse the filesystem once when deglobbing

Fuuzetsu:
For warnings, we simply recover which patterns haven't matched any of
the files that were found to match with globDir.


sol:
That is pretty neat!

sol:
Turns out it's even simpler https://github.com/sol/hpack/commit/8288913c39988602d244bac5390822ebac188d24

sol:
Well and https://github.com/sol/hpack/commit/af484b059fca49d17044a0991fcd2a4f6e01f2aa

Issue #40, opened 2015-07-15
support extensions, tested-with, benchmarks


simonmichael:
hpack doesn't recognise `extensions:` or `tested-with:`. Also it should allow benchmarks to be specified (just like tests).

sol:
@simonmichael Firstly, a big thanks for giving feedback on `hpack`.  That is really useful!

We want support for `benchmarks`, and implementing this should be easy and very similar to what we already have for `executables` and `tests`.  @simonmichael I'm not sure when I get around to do this, but if you send a PR I'll make sure it hits Hackage soon.

As for `extensions`, I don't have the full picture here.  I know that Cabal allows you to list them, but don't know how they are used or what the purpose is.  If we want them in the cabal file, I guess we could infer them from the source files?  @simonmichael any input on this?  Listing them twice, in source files and in the `package.yaml` is violating the design goals of `hpack` (that is DRYing things up).  That said, I may still agree to add it, if somebody makes the case for it.  (btw, `default-extensions` are already supported).

As for `tested-with`, for my packages I don't want to maintain that manually.  I think ideally Hackage would try to build uploaded packages with multiple versions of GHC and add that information to the website.  What I currently do is maintain that information in my `.travis.yml` and/or I use version constraints on `base` to communicate with which versions of GHC a package works.  All that said, I'm happy to accept a patch for `tested-with` if there is demand for it ;).

simonmichael:
Ok, I assumed it was a goal to support all valid cabal file constructs, but maybe not. I must admit part of me likes the simplicity of flagless cabal files and hopes #30 never lands. :-)

Re testing buildability, the hackage matrix builder does that: http://matrix.hackage.haskell.org/package/hledger

sol:
Don't get me wrong.  If if somebody wants to use `hpack` but the only thing holding him back is missing support for `extensions` or `tested-with`, then let's just add it.  I just want to make sure that somebody really, really wants this, and does not only e.g. list `extensions` because Cabal has support for it.

My point is that I think `Cabal` and `cabal-install` are not perfect.  This is the reasons why tools like [sol/hpack](https://github.com/sol/hpack), [zalora/tinc](https://github.com/zalora/tinc) and [commercialhaskell/stack](https://github.com/commercialhaskell/stack) exist in the first place.  For that reason I don't take Cabal as the authoritative source of truth.

[...]

PR #43, opened 2015-08-13
tested-with

phadej:
Related to https://github.com/sol/hpack/issues/40

`tested-with` is useful, as one could [generate travis build matrix](https://github.com/hvr/multi-ghc-travis/blob/master/make_travis_yml.hs) using that information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /dist/
+.stack-work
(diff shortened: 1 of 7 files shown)


qoelet:
We skipped the `stack` related stuff for now, rebased and squashed the commits and pushed it to master.

Thanks a lot for the contribution!

PR #51, opened 2015-09-09
Add benchmarks

### LINKED ISSUE #40: support extensions, tested-with, benchmarks

phadej:
Partially resolves #40

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -50,7 +50,7 @@ import           Hpack.Util
 import           Hpack.Yaml
 
 package :: String -> String -> Package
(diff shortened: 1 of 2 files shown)


phadej:
example:
- [package.yaml](https://github.com/phadej/binary-tagged/blob/master/package.yaml)
- [generated .cabal](https://github.com/phadej/binary-tagged/blob/master/binary-tagged.cabal)

sol:
@phadej Sorry for the late reply.  LGTM.  I'll push this to Hackage later today.

phadej:
One week is totally acceptable :) Thanks!

Issue #53, opened 2015-09-22
hpack-0.8.0 test suite failure


peti:
Citing from http://hydra.cryp.to/build/1306557/nixlog/2/raw:

~~~~
Failures:

  test/Hpack/ConfigSpec.hs:64:
  1) Hpack.Config.parseJSON, when parsing a Dependency, when parsing fails, returns an error message
       expected: Left "Error in $: expected String or an Object, encountered Number"
        but got: Left "when expecting a String or an Object, encountered Number instead"

  test/Hpack/ConfigSpec.hs:69:
  2) Hpack.Config.parseJSON, when parsing a Dependency, when parsing fails, when ref is missing, produces accurate error messages
       expected: Left "Error in $: key \"ref\" not present"
        but got: Left "key \"ref\" not present"

  test/Hpack/ConfigSpec.hs:78:
  3) Hpack.Config.parseJSON, when parsing a Dependency, when parsing fails, when both git and github are missing, produces accurate error messages
       expected: Left "Error in $: neither key \"git\" nor key \"github\" present"
        but got: Left "neither key \"git\" nor key \"github\" present"

  test/Hpack/ConfigSpec.hs:643:
  4) Hpack.Config.readPackageConfig, when package.yaml is invalid, returns an error
       expected: Left "package.yaml: Error in $.executables.foo: failed to parse field executables: The key \"main\" was not found"
        but got: Left "package.yaml: The key \"main\" was not found"

  test/Hpack/UtilSpec.hs:71:
  5) Hpack.Util.List, when parsing single values, returns error messages from element parsing
       expected: Left "Error in $: neither key \"git\" nor key \"github\" present"
        but got: Left "neither key \"git\" nor key \"github\" present"

  test/Hpack/UtilSpec.hs:78:
  6) Hpack.Util.List, when parsing a list of values, propagates parse error messages of invalid elements
       expected: Left "Error in $[1]: neither key \"git\" nor key \"github\" present"
        but got: Left "neither key \"git\" nor key \"github\" present"
~~~~

peti:
Ping?

soenkehahn:
This could be fixed by specifying `aeson >= 0.10`. But would that help in the nix case?

peti:
NixOS follows Stackage, and LTS Haskell specifies aeson-0.8.0.2. Even Stackage Nightly is still at aeson-0.9.0.1.

sol:
[...]

PR #56, opened 2015-10-19
Update .travis.yml

phadej:
Also resolves: https://github.com/sol/hpack/issues/53

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,90 @@
-language: haskell
-ghc: 7.8
+# This file has been generated -- see https://github.com/hvr/multi-ghc-travis
+language: c
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.cabsnap
+    - $HOME/.cabal/packages
+
+before_cache:
+  - rm -fv $HOME/.cabal/packages/hackage.haskell.org/build-reports.log
+  - rm -fv $HOME/.cabal/packages/hackage.haskell.org/00-index.tar
+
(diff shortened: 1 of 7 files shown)


sol:
@phadej Again, sorry for the delay on this.

Firstly, thanks for taking the time to look into this.  I understand that the current situation is not ideal and causes friction with Stackage and Nix.  But I still assume that this is temporary and will eventually resolve itself.  For that reason I'm inclined to solve this by applying #55 instead, mainly for the reason that it keeps the code simple.

For Stackage and Nix hope that disabling the tests provide a temporary solution until they are on the latest version of `aeson`.

sol:
Regarding the travis changes, I like the idea of generating `.travis.yml`.  If we could make that work with [`zalora/tinc`](https://github.com/zalora/tinc) that could be quite useful (see https://github.com/sol/hpack/commit/1e9afffdee6cdf02c51f97937f6262c44ce436bf).

phadej:
hvr's `multi-ghc-travis` script isn't very customisable (in fact in this PR I modified it by hand after generation). In my own projects, I use [`phadej/travis-meta-yaml`](https://github.com/phadej/travis-meta-yaml) but I don't know how usable it is for others atm.

sol:
@phadej I'm inclined to close this one.  Is that ok with you?

phadej:
Please close, I made a new iteration: https://github.com/sol/hpack/pull/72 :)

PR #57, opened 2015-11-03
Support other-extensions

phadej:


package.yaml
```diff
@@ -26,6 +26,27 @@ library:
...
+    - FlexibleInstances
+    - OverloadedStrings
+    - PolyKinds
+    - QuasiQuotes
+    - RecordWildCards
+    - ScopedTypeVariables
+    - TypeFamilies
+    - TypeOperators
+    - UndecidableInstances
+    - ViewPatterns
```
sol: Unless somebody can point out how this information is essential, I'm not eager to maintain this list manually.

As I understand it, it's possible to derive this list of extensions by looking at the source files.  If we want to have this information in the cabal file for completeness, I guess `hpack` could infer it by looking at the source files.

sol:
@phadej So far I failed to see the practical benefit of `other-extensions`, but I may have easily missed something here.  Do you have a deeper insight?

If we come to the conclusion they are useful, let's infer them.  WDYT?

phadej:
@sol `cabal-install` does use that information, i.e. if compiler doesn't support extensions mentioned there, it doesn't even bother to build the package.

Arguably the benefit isn't big. But I'd like `hpack` to be as little opinionated as possible. 

And as a side comment, with haskell-src-exts one could generate the `other-extensions` field in simple cases (no CPP'd extensions) - that I could have as an external utility working on package.yaml though.

sol:
I'm puzzled whether putting a constraint on `base` isn't a more obvious way to achieve the same goal?

When it comes to being opinionated: I want `hpack` to be as useful as possible for as many people as possible.  But looking at Cabal, I think there some features in Cabal that don't make sense.  This is why I wrote `hpack` in the first place.

So if `hpack` tries to be better than Cabal, is being opinionated to some degree a healthy approach?

That said, if there is broad demand for this, I'm probably willing to add it with the assumption that it is not really harmful.

Regarding `haskell-src-exts`, it takes a long time to build.  I think it should be relatively easy to extract language pragmas with custom code.  I would assume, we have to take the superset from all CPP branches, otherwise the generated cabal file depends on the users environment.  Does this make assumption make sense?

Replicating the conditionals, as suggest in https://github.com/phadej/cabal/commit/6916f3b53db19d4442914f4473e21669114025ba, personally I'm not sure if I would be willing to do that for my projects, as you would need to maintain the invariants manually.  Inferring them may be none-trivial.

phadej:
Regarding the conditionals, I'd omit any pragmas inside CPP conditionals; hiding them behind the flag is *almost* only for informational purposes.

I don't want *strictly* restrict all of my package to be GHC-only. Some of them could be easily compiled even with `Haskell98` + trivial extra compatible compiler. Not that I tried. But I'll try my best to help any who is willing to do that.

Indeed `haskell-src-exts` is heavy, I could work on custom code too, but then I'd need a bit more time.

sol:
@phadej I merged your patch with the assumption that it is not really harmful to have it.  For my own projects I'm not going to maintain that information, for the reason that I think the cost / benefit ratio is not high enough (that includes `hpack` itself, see https://github.com/sol/hpack/commit/897e840d87ad215194fb1163bb3a9efc443cd704).

Sorry for the delay on this.

phadej:
@sol, it's ok for me. I added them to `hpack` as an example.

PR #36, opened 2015-07-11
add support for conditionals

sol:

Issue #29, opened 2015-07-06
Support for os-specific things (dependencies, etc.)


sol:
We still need to settle on some concrete syntax here.  Ideas?

sol:
Ideally the approach can be extended to `arch` and `flag`, if we ever reach the point that we need it.

Relevant Cabal documentation: https://www.haskell.org/cabal/users-guide/developing-packages.html#conditional-blocks

cc @Fuuzetsu @soenkehahn

sol:
Some ideas:

```yaml
dependencies:
  - name: Win32
    when: os(windows) && arch(i386)
  - name: unix
    when: "!os(windows)"

windows-i386-dependencies:
  - Win32
unix-dependencies:
  - unix
```

```yaml
os:
  windows:
    arch:
      i386:
        dependencies: Win32
  other:
    dependencies: unix
```

```yaml
if:
  condition: os(windows) && arch(i386)
  then:
    dependencies:
      - Win32
  else:
    dependencies:
      - unix
```

sol:
```yaml
when:
  condition: os(windows) && arch(i386)
  dependencies:
    - Win32
when:
  condition: "!os(windows)"
  dependencies:
    - unix
```

sol:
when:
  os: windows
  arch: i386
  dependencies:
    - Win32

when:
  os: "!windows"
  dependencies:
    - unix

sol:
```yaml
when:
  condition: os(windows) && arch(i386)
  then:
    dependencies:
      - Win32
  else:
    dependencies:
      - unix
```

sol:
For flags:

```yaml
flags:
  integration-tests:
    description: Run the integration test suite
    manual: yes
    default: no
```
or
```yaml
flags:
  - name: integration-tests:
    description: Run the integration test suite
    manual: yes
    default: no
```

qoelet:
```yaml
when:
  - condition: os(windows) && arch(i386)
    dependencies:
      - Win32
  - condition: "!os(windows)"
    dependencies:
      - unix
```

jkarni:
Any decisions? The last one, and either of the flag proposals, sound good to me. Happy to help implementing.

sol:
Yes, sounds good to me. Maybe the first flags proposal?

phadej:
I long avoided need to flags, but atm I just cannot avoid it anymore. So I'd like to see this happen. I'd like the
```yaml
flags:
  integration-tests:
    description: Run the integration test suite
[...]

PR #60, opened 2016-01-13
Conditionals and flags

phadej:
Supersedes https://github.com/sol/hpack/pull/36
Implements https://github.com/sol/hpack/issues/29

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /dist/
+.stack-work/
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -32,6 +32,7 @@ library
     , Glob
     , text
     , unordered-containers
+    , vector
     , yaml
     , aeson >= 0.8
   exposed-modules:
@@ -62,6 +63,7 @@ executable hpack
     , Glob
     , text
     , unordered-containers
+    , vector
     , yaml
     , hpack
     , aeson >= 0.8
@@ -85,6 +87,7 @@ test-suite spec
     , Glob
     , text
     , unordered-containers
+    , vector
     , yaml
     , hspec == 2.*
     , temporary
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -25,6 +25,7 @@ library
   build-depends:
       base >= 4.7 && < 5
     , base-compat >= 0.8
+    , bytestring
     , deepseq
     , directory
     , filepath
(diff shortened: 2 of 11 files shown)


phadej:
Travis has a bad day today. Can someone restart job later?

mgsloan:
+1

Perhaps we can consider different defaults?  IMHO, `FlagFields { _flagFieldManual = True, _flagFieldDefault = False, _flagFieldDescription = Nothing }` seems like a better default.  In other words, by default, you need to ask for the flag in order for it to be applied.  It goes against the grain of cabal, but it seems like the point of hpack is making the recommended / standard thing be concise.

phadej:
@mgsloan I guess which flags *defaults* you prefer is dictated by your primary use. To make conditional dependencies `manual = False, default = True` is handy; OTOH for enabling optional features (like examples / executables) `manual = True, default = False` is better.

Alternatively one could make `manual` and `default` mandatory.

I don't have strong opinion on this though.

soenkehahn:
I'm very much in favor of `manual = True`.

phadej:
ping @sol and @qoelet

phadej:
The feature still missing is handling of:
```yaml
library:
  source-dirs: src
  when:
    - condition: impl(ghc >= 7.8)
      exposed-modules:
        - Data.Aeson.Extra.SingObject
        - Data.Aeson.Extra.SymTag
```

But that would require bigger changes, so I rather do it in another PR.

phadej:
I don't know if tinc could execute cabal with a bit more backjumps...

phadej:
ping @sol, @soenkehahn, @mgsloan

Rebased, removed `manual` and `default` defaults, which is probably better as the defaults aren't obvious.

sol:
A problem with my initial code is that it does not handle warnings in conditional blocks.

phadej:
@sol, I see. I'll fix that.

phadej:
@sol, It's fixed. The approach is a bit unsatisfactory as the framework highly assumes we parse flat data.

I could make largish refactor, introducing `StrictFromJSON a` which would produce `Parser (CaptureUnknownFields a)` directly.

sol:
@phadej I think this still has issues.  E.g. it does not warn on `github` inside a top-level conditional block, but `github` is never allowed in a conditional block.

sol:
I gave conditionals an other try #78.

sol:
@phadej Let's sort out #78 first.  Once we have support for conditionals on `master` let's tackle support for flags.

PR #66, opened 2016-02-20
Resolve #65

### LINKED ISSUE #65: Test fails on Mac OSX
```
  2) Hpack.Config.getModules, when source directory is './.', ignores Setup
       expected: ["Foo"]
        but got: ["Foo","Setup"]
```

this because other dir in path is `/private/var/...` and other `/var/...` - i.e. `src == dir` check isn't enough. One would need to follow links

```
$ ls -l /
...
lrwxr-xr-x@   1 root  wheel    11 Oct 23  2014 var -> private/var
```

phadej:


sol:
I think the actual issue is in `readPackageConfig`.  Here we already call `canonicalizePath`

```haskell
readPackageConfig file = do
  ...
      dir <- takeDirectory <$> canonicalizePath file
```

I guess this has to be changed to

```haskell
readPackageConfig file = do
  ...
      dir <- canonicalizePath $ takeDirectory file
```

sol:
In addition, let's enable travis for Mac OS X.

sol:
@phadej I propose #70 instead.  Can you try this and give me your feedback?

sol:
BTW, given the complexity of adapting `.travis.yml` for OS X, I'm not sure anymore if we really want this.  In the past I didn't discover many discrepancies between Linux and OS X.  So I'm not sure if it's worth the additional effort.

sol:
Fixed by applying #70.

PR #72, opened 2016-02-21
Build with stack also

phadej:


phadej:
osx obviously fails.

sol:
@phadej Builds on OS X seem to take a long time.  Are build artifacts properly cached?

Issue #73, opened 2016-02-21
Allow specifying license file


tfausak:
My license files typically have an extension. hpack only looks for a file called `LICENSE` with no extension. I tried to manually set the `license-file`, but that is an unknown field. I would like to be able to manually set the license file. It would be nice if license files with extensions could automatically be found.

mitchellwrosen:
I implemented this in #74 but I didn't include any automatic discovery.

tfausak:
Fixed by #74. Thanks!

PR #74, opened 2016-02-22
Allow license-file to be specified

mitchellwrosen:
#73

### DIFF
--- a/test/Hpack/ConfigSpec.hs
+++ b/test/Hpack/ConfigSpec.hs
@@ -258,6 +258,15 @@ spec = do
         )
         (`shouldBe` package {packageLicenseFile = Just "LICENSE"})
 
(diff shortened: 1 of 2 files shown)


tfausak:
:+1:

test/Hpack/ConfigSpec.hs
```diff
@@ -258,6 +258,15 @@ spec = do
         )
         (`shouldBe` package {packageLicenseFile = Just "LICENSE"})
 
+    it "accepts license file" $ do
+      withPackageConfig [i|
+        license-file: FOO
+        |]
+        (do
+        touch "FOO"
```
sol: In cases where the user specifies the license file explicitly I would always include it, without any existence checks.

sol:
Thanks for the contribution!

Issue #71, opened 2016-02-21
Add CI for Windows


sol:
(probably using ci.appveyor.com)

@phadej Are you interested to look into this?

phadej:
I can check this next week.

phadej:
Setting up the appveyor is trivial, the tests aren't:

```
  test\Hpack\ConfigSpec.hs:282: 
  1) Hpack.Config.readPackageConfig accepts data-files
       Actual list is not a permutation of expected list!
         expected list contains:   ["data/foo/index.html", "data/bar/index.html"]
         actual list contains:     ["data\\bar\\index.html", "data\\foo\\index.html"]
         the missing elements are: ["data/foo/index.html", "data/bar/index.html"]
         the extra elements are:   ["data\\bar\\index.html", "data\\foo\\index.html"]
 
  test\Hpack\UtilSpec.hs:173: 
  2) Hpack.Util.expandGlobs, when expanding *, expands by extension
       expected: ([],["files/bar.js","files/baz.js","files/foo.js"])
        but got: ([],["files\\bar.js","files\\baz.js","files\\foo.js"])
 
  test\Hpack\UtilSpec.hs:178: 
  3) Hpack.Util.expandGlobs, when expanding *, rejects dot-files
       expected: ([],["foo/bar"])
        but got: ([],["foo\\bar"])
 
  test\Hpack\UtilSpec.hs:183: 
  4) Hpack.Util.expandGlobs, when expanding *, accepts dot-files when explicitly asked to
       expected: ([],["foo/.baz"])
        but got: ([],["foo\\.baz"])
 
  test\Hpack\UtilSpec.hs:188: 
  5) Hpack.Util.expandGlobs, when expanding *, matches at most one directory component
       expected: ([],["foo/bar.js"])
        but got: ([],["foo\\bar.js"])
 
  test\Hpack\UtilSpec.hs:194: 
  6) Hpack.Util.expandGlobs, when expanding **, matches arbitrary many directory components
       expected: ([],["foo/bar/baz.js"])
        but got: ([],["foo\\bar\\baz.js"])
 
Randomized with seed 195582947
```

sol:
All of these should be easy to fix.  But as we have `hpack` support in `stack` now, this is kind of important.

@phadej Can you open a PR with your AppVeyor config?

sol:
Build for windows is passing now.

PR #76, opened 2016-02-26
Fix build for Windows (see #71)

sol:


sol:
@phadej I hop this fixes the build for Windows, but as I don't have a Windows box available, we have to wait for the build.

PR #81, opened 2016-03-01
Add support for flags

sol:


sol:
@phadej I started from 087395f, but end up refactoring it quite a bit.

phadej:
You made it earn about unknown fields. Great. LGTM

PR #83, opened 2016-03-02
Add ld-options and buildable flag

sol:


sol:
See https://github.com/commercialhaskell/stack/pull/1860.

PR #88, opened 2016-04-01
Only re-generate cabal file if hpack is new enough

sol:
That is: If a user tries to re-generate an existing cabal file with an 
older version of `hpack` than the one that was used to generate the 
existing cabal file, then do nothing.

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -6,6 +6,8 @@ module Hpack (
 #ifdef TEST
 , hpackWithVersion
 , parseVerbosity
+, extractVersion
+, parseVersion
 #endif
 ) where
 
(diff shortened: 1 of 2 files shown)


sol:
@mgsloan @phadej This might be a more convenient fix for https://github.com/commercialhaskell/stack/issues/1813 / https://github.com/commercialhaskell/stack/pull/1815.

Instead of relying on timestamps, you can just bump the version of your custom version of `hpack` to e.g. `99.99.99`, and upstream `hpack` will never re-generate your cabal file.

mgsloan:
Seems reasonable, sticking with timestamps for now seems good too until there's some particular issue with that

sol:
@mgsloan Yes, I agree that we should stick with the timestamps for now.

PR #89, opened 2016-04-11
Allow custom file name instead of package.yaml

rvion:
@sol 

Following https://github.com/commercialhaskell/stack/issues/2011
and https://github.com/commercialhaskell/stack/issues/2001

hpack is great, but I was a bit annoyed in my projects because

  - stack access several time a same cabal file for some operations => hpack is run too often for me
  - there is too much logging by default (dozen of line per rebuild when many subprojects) but in the same time, I'm not sure less logging is better.
  - I have dozen of files package.yaml in my editor file tree (because all subproject have one) and it's a bit more annoying to find the right one than if I had files with different names.
  - stack maintainer don't really value adding a disable hpack flag (which I totally understand)
  - I cant move easilly my package description file elsewhere because of hpack internal (and I don't really want to because of folder organisation)

So, I made a small patch that allow one to specify a custom package description name instead of using the default "package.yaml" name. This way, I can solve all my problems: stack don't know anymore I'm using hpack, I can run hpack only when I want (with something like that: https://atom.io/packages/save-commands, it's rather painless)

implementation note:
I tried not to change the API used by stack, so no modification have to done on the stack project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /dist/
+.stack-work/
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -34,9 +34,9 @@ import           Hpack.Run
 programVersion :: Version -> String
 programVersion v = "hpack version " ++ Version.showVersion v
 
-header :: Version -> String
-header v = unlines [
-    "-- This file has been generated from " ++ packageConfig ++ " by " ++ programVersion v ++ "."
+header :: Version -> FileName -> String
+header v file = unlines [
+    "-- This file has been generated from " ++ file ++ " by " ++ programVersion v ++ "."
   , "--"
   , "-- see: https://github.com/sol/hpack"
   , ""
@@ -49,6 +49,7 @@ main = do
     ["--version"] -> putStrLn (programVersion version)
     ["--help"] -> printHelp
     _ -> case parseVerbosity args of
+      (verbose, [dir, fname]) -> hpackWithName fname dir verbose
       (verbose, [dir]) -> hpack dir verbose
       (verbose, []) -> hpack "" verbose
       _ -> do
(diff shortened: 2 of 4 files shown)


sol:
Other tools, e.g. `tinc` assume `package.yaml`.  For that reason I'm not excited about the idea to make it customizable.

mgsloan:
What if it was like `pkg-name.hpack` like cabal files?  Could even have this format of file imply the package name.  It is unfortunate that editors won't know to give yaml highlighting, but hey.

rvion:
@sol I understand your concern, and have no strong opinion on that (I was not aware of `tinc`).

quick notes: 

 - `package.yaml` is still the default name hpack look for unless a custom name is specified
 -  @mgsloan I personally use pkg-name.hpack.yaml :)

soenkehahn:
@mgsloan: I really like the fact that `package.yaml` files -- unlike cabal files -- all have exactly the same name. It just makes a few things a little easier, e.g.

- finding the file in a list of alphabetically sorted files (because you don't have to remember the project name to know where to look),
- grepping through multiple `package.yaml` files in different projects, etc.

Also, it very naturally enforces the constraint that you should only have one `package.yaml` file in your project.

I never understood why cabal files start with the project name. Seems redundant to me and I don't see any advantages.

mgsloan:
Yup, seems reasonable to me!

rvion:
@soenkehahn it's all about habits and usages, there is no superior design I think. 

 - I mostly don't use a tree view in my editors, and rely on names to fuzzy find files I want.
 - xxx.hpack.yaml allow to grep through multiple package specs in different folders, etc, and I can disambiguate them via filenames instead of path.

this being said, I agree that having one same name is easier for tooling, and I'll even probably switch back to use `package.yaml` because it doesn't change much in the end in the end.

thanks for the discussion

PR #92, opened 2016-04-28
Fix build on ghc-8

mgsloan:
Due to addition of https://ghc.haskell.org/trac/ghc/ticket/10030 , which is ambiguous with `Hpack.Config.packageName`


mgsloan:
Travis failure seems to be due to hackage availability

sol:
Build green now.  Thanks a lot!

PR #95, opened 2016-05-04
Add 'hpackSilent', with result datatype

mgsloan:
See https://github.com/commercialhaskell/stack/issues/2011#issuecomment-216732959


mgsloan:
Pushed an amended version which avoids API changes and fixes the test-suite build

PR #98, opened 2016-05-09
Rename hpackSilent to hpackResult for consistency

robbinch:


robbinch:
@mgsloan For consistency with eg. `quickCheckResult` and `hspecResult`

mgsloan:
Ahh, yeah, I like that naming much better.

PR #110, opened 2016-06-09
test/Helper.hs: fix build with GHC 8.0.1

borsboom:
It'd be great to do a minor release to Hackage with this change so that this package can be added back to Stackage.  This is currently the only thing blocking adding `stack` back into Stackage since the GHC 8.0.1 upgrade.


soenkehahn:
Fixed this in 2b1a38a753d8ca6aa3716cea7c611bc6a0c33fc7 and released to hackage as `0.14.1`. Thanks for your contribution!

PR #114, opened 2016-08-02
Add conversion from cabal files

### LINKED ISSUE #63: Automatic generation of hpack from .cabal files?
Hpack looks really nice, great work!  I think a good step towards adoption would be to support automatic conversion of cabal files to hpack format.  This could be done in a new package, `hpack-convert`, or something, to avoid a dependency on Cabal in the main hpack library (maybe that dependency would be ok, though?).

yamadapc:
Adds modules `Hpack.Convert` and `Hpack.Convert.Run` for reading a
`GenericPackageDescription` onto a `Package` and running the user-facing
conversion routine.

The user-facing option is exposed through:
```
$ hpack --convert [cabal-file|project-dir]
```

- - -

Also adds tests testing we can read cabal files generated from some of
the README examples and hpack's into package.yamls that look like the
original ones.

Everything should be supported, including converting conditionals.

I really need feedback on how to make `Hpack.Config` `Package` related
`ToJSON` instances cleaner. That's the ugly bit of the code.

There some cases where we run into problems in conversion. For example,
the Cabal library does not expose in its data-structures if `buildable`
was specified or not. To keep the output clean, I initially striped it
from sections if it was true, but then empty conditional blocks could be
generated. Right now it'll only output `buildable` if it's not
redundant based on context. It'll also remove any empty conditional
branches.

This closes #63.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -6,13 +6,16 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
 module Hpack.Config (
   packageConfig
 , readPackageConfig
+, encodePackage
+, writePackage
 , renamePackage
 , packageDependencies
 , package
@@ -41,17 +44,23 @@ module Hpack.Config (
 
 import           Control.Applicative
 import           Control.Monad.Compat
+import           Data.Aeson
 import           Data.Aeson.Types
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString as ByteString
 import           Data.Data
 import           Data.Map.Lazy (Map)
 import qualified Data.Map.Lazy as Map
 import qualified Data.HashMap.Lazy as HashMap
-import           Data.List.Compat (nub, (\\), sortBy)
+import           Data.List.Compat (elemIndex, intersect, isPrefixOf, nub,
+                                   sortBy, (\\))
 import           Data.Maybe
 import           Data.Ord
 import           Data.String
 import           Data.Text (Text)
 import qualified Data.Text as T
+import qualified Data.Vector as Vector
+import qualified Data.Yaml.Pretty as Yaml
 import           GHC.Generics (Generic, Rep)
 import           Prelude ()
 import           Prelude.Compat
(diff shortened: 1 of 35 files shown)


yamadapc:
Tests are failing because of a key sorting and an encoding issue. I'll fix later today.

mgsloan:
Awesome!!

src/Hpack/Config.hs
```diff
@@ -270,6 +301,252 @@ data PackageConfig = PackageConfig {
...
+                  let srepo = (sourceRepositoryUrl repo) ++
+                          (fromMaybe "" (sourceRepositorySubdir repo))
+                  in
+                  if githubBaseUrl `isPrefixOf` srepo
+                     then let dropIfGH (Just (String v))
+                                  | githubBaseUrl `isPrefixOf` T.unpack v = Nothing
+                              dropIfGH v = v
+                          in
+                          HashMap.insert "github" (String (T.pack (drop (length githubBaseUrl) srepo))) $
+                          HashMap.alter dropIfGH "bug-reports" $
```
sol: If it's me, I would try to avoid this kind of logic here and instead do things like this in a conversion step.

sol:
Hey, this looks really useful. But it's quite some code and my bandwidth is still limited. @yamadapc would you be willing to maintain this as a separate package?

We would then add prominent references to the README etc., so that discoverability is not a problem.

I think even if I would do it myself, I would at least go for two separate executables (UNIX style).

yamadapc:
I'm fine with doing that.

I started it and ran into a couple of problems:
- I need to expose `HasFieldNames` from `Hpack.Config`
- I need to expose `Hpack.GenericsUtil` and `Hpack.Util`

Those are the ones I can solve with a fork/PR, no biggie.

- - -

**And then there's one which I can't understand**:
```
        Couldn't match type ‘Rep SourceRepository’
                       with ‘GHC.Generics.M1 GHC.Generics.D d1 m1’
        The type variables ‘d1’, ‘m1’ are ambiguous
        In the expression: genericToJSON_
        In an equation for ‘toJSON’: toJSON = genericToJSON_
        In the instance declaration for ‘ToJSON SourceRepository’
```

In all uses of the `genericToJSON_` function. I think it has to do with the `typeName` stuff only working for data types defined in the current module. I'm not familiar with `GHC.Generics`. 

[Maybe it's what these guys are talking about.](http://stackoverflow.com/questions/32226158/what-does-this-ghc-generic-function-signature-error-message-mean).

Ideas?

soenkehahn:
@yamadapc: Do you have a `wip` branch where we can take a look at the code producing that `Generics` type error?

yamadapc:
@soenkehahn This gives you the error: https://github.com/yamadapc/hpack-convert-separate-package-with-hpack-dependency

But talking with Simon, we had reached the conclusion that it'd be easier to leave `hpack-convert` on a separate package that'd just fork `hpack`. I just need to do the laundry on this `enhancement/63-convert-cabal-files` branch and update the executable/README.

soenkehahn:
@yamadapc: I see. Looking forward to this. :)

yamadapc:
Ok, it has been published on Hackage as `hpack-convert` and is on my GitHub account as an `hpack` fork in https://github.com/yamadapc/hpack-convert.

People can choose to:
- [Install from source](https://github.com/yamadapc/hpack-convert#installing-from-source)
- [Download a binary](https://github.com/yamadapc/hpack-convert#download-a-pre-built-binary)
- ``curl -F "cabalfile=@./`echo *.cabal`" https://hpack-convert.herokuapp.com``

I'll keep it on my todo list to make it clean and have it merged back upstream. Even though it's a fork it's not hard to just cherry-pick commits back here at any point.

- - -

Closing this in favor of this approach.

sol:
We may still want to add it to the README. PRs welcome!

soenkehahn:
@yamadapc: Just tried this out, awesome work!

PR #117, opened 2016-08-10
"library fields" -> "library field" in readme

mitchellwrosen:


mitchellwrosen:
The CI failure wasn't due to these changes, not sure what's going on there

sol:
Is this really better than before?  We can only have one library section, but that section can contain multiple "fields".

Let me know if you still think we should merge this.

mitchellwrosen:
Oh, you're right. I was confusing "field" with "section".

PR #120, opened 2016-08-12
Fix rendering dots in multi-line fields

### LINKED ISSUE #119: Problems when a directory field contains a dot
When a field that takes a directory contains a dot, hpack's renderer will generate a line with nothing but a `.` in it. However, this is invalid in a cabal file, as dots on their single lines are meant to express empty-lines (I think, I can't find the documentation [the source that does it is here](https://github.com/haskell/cabal/blob/9b4712583ca390d0f034cc66eca4de51c00cc8d0/Cabal/Distribution/ParseUtils.hs#L576)).

A simple example of the bug is here: https://gist.github.com/0cc7f8f8e2cbc9ee7bbf82876a822e28

In there, `source-dirs: .` generates an invalid cabal file. Though this specific case isn't super common, without specifying `source-dirs: .`, `hpack` won't include the `other-modules` section for the modules in the current directory (I'd consider that another ticket).

- - -

My idea for a fix would be to output:
```
field: .
  , other-value
  , other-value
```

Whenever `.` is a value.

yamadapc:
This closes #119. By pulling dots to the same line as the field name.

For example:

    source-dirs:
    - .
    - something

Will render as:

    hs-source-dirs: .
                 , something

Or

    hs-source-dirs: .,
                   something

Depending on the comma configuration. Likewise:

    include-dirs:
    - .
    - something

Will render as:

    include-dirs: .
          something

### DIFF
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -33,7 +33,7 @@ library
     , containers
     , unordered-containers
     , yaml
-    , aeson >= 0.11
+    , aeson >= 0.8
   exposed-modules:
       Hpack
       Hpack.Config
--- a/src/Hpack/Render.hs
+++ b/src/Hpack/Render.hs
@@ -40,7 +40,7 @@ data Value =
 data Element = Stanza String [Element] | Group Element Element | Field String Value
   deriving (Eq, Show)
 
-data Lines = SingleLine String | MultipleLines [String]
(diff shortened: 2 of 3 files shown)


yamadapc:
Closed in favor of #125.

Issue #49, opened 2015-09-08
Support for "build-type: Custom"


peti:
Is it possible to specify `build-type: Custom` somehow? I tried adding an appropriate entry to my `package.yaml` file, but `hpack` complained about an unsupported entry.

mitchellwrosen:
`build-type: Simple` is hard-coded

sol:
There is no support for `build-type: Custom` and it's unlikely to happen.  This feature would tie `hpack` to `Cabal` and would make it hard to implement alternative consumers of `package.yaml` files.

peti:
The problem I have is that it's not possible to run a `doctest` test suite reliably in a "Simple" build: https://github.com/fpco/stackage/issues/1616. Or, well, it is possible but it requires manual maintenance of a list of dependencies, which isn't exactly pleasant.

sol:
After discussing this again with @soenkehahn I'm willing to compromise on my original goal for the sake of making `hpack` as useful as possible.

To that end I am willing to accept a patch for `build-type`.

PRs please!

sol:
This is released by now.

PR #126, opened 2016-08-25
Please support Setup.hs

osbugs:
Especially in complex project setups where a package configuration with hpack is a big benefit, it is highly likely that the build process is tuned with a custom build type (to be found in Setup.hs).

This patch tries to figure out whether such a custom build setup is in place, and sets the Cabal entry accordingly.

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -35,7 +35,7 @@ import           Hpack.Config
 import           Hpack.Run
 
 programVersion :: Version -> String
-programVersion v = "hpack version " ++ Version.showVersion v
+programVersion v = "hpack (custom) version " ++ Version.showVersion v
 
 header :: Version -> String
 header v = unlines [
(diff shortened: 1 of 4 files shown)


mitchellwrosen:
This was discussed a bit in #49. @sol said:

> There is no support for build-type: Custom and it's unlikely to happen. This feature would tie hpack to Cabal and would make it hard to implement alternative consumers of package.yaml files.

osbugs:
I see your point. Luckily, the patch does not feature a new "build-type:" entry in the package.yaml but instead looks for a Setup.hs file. In contrast to the build-type entry, whose value has Cabal-specific meaning ("Custom", "Simple"), the Setup.hs file has nothing to do with Cabal, and any build system can call it. What you do inside may or may not be Cabal-specific.

sol:
The presence of `Setup.hs` does not necessarily imply `build-type: Custom`, but I'm willing to accept a patch for `build-type` (see https://github.com/sol/hpack/issues/49#issuecomment-243018166).

sol:
closing this

PR #127, opened 2016-09-02
Support for non-Simple build-type

mitchellwrosen:
#49


sol:
Awesome, looks good to me.

One minor thing, can you please split the tests into multiple cases.

"accepts build-type Simple" etc.

One expectation per test.

mitchellwrosen:
Yup! I always seem to mess up the test suite somehow. Done and force pushed.

PR #131, opened 2016-09-11
Support globs in c-sources

sol:


sol:
I applied some refactoring on top of your changes. To see the total diff you may want to look at #131.

PR #141, opened 2017-01-01
Add more documentation to declaring conditionals.

deech:


README.md
```diff
@@ -134,6 +134,49 @@ becomes
...
+becomes:
+
+    if os(linux)
+      extra-lib-dirs:
+        /usr/lib/x64
+
+Unlike with `.cabal` files compound predicates need to be in parens, for example:
+
+    when:
+      - condition: (!os(darwin) && !os(windows))
```
sol: I believe the issue here is not that you have "compound predicate" here.  It's merely the `!`, which in YAML must be quoted when at the beginning of a string, say:

    condition: "!os(darwin) && !os(windows)"

README.md
```diff
@@ -134,6 +134,49 @@ becomes
...
+
+    if os(darwin)
+      extra-lib-dirs:
+        lib/darwin
+    if os(linux)
+      extra-lib-dirs:
+        /usr/lib/x64
+
+If conditionals are declared under different `when` clauses only the last one
+will make it to the `.cabal` file:
```
sol: I think it may not necessarily the last one.  This is again resulting from how YAML works.  This is an object (think of it as a hash map), duplicate keys are not allowed.

sol:
@deech First thing first, hanks for taking the time to open a PR.

As it currently stands, it's more like "this surprised me, and this is the workaround".  But those things are not really surprising, they result from how YAML works.

I'm not suggesting we should not address this.  But, can we address this in a way so that the user gets an understanding of the underlying mechanics (YAML, in this case)?

deech:
Hi,
Thanks for hpack! I loveit.

Having read your comments I agree that this is a YAML and not an hpack issue. I'm obviously not as familiar with YAML but as a Cabal user it did trip me up for a few hours. How do you feel about amending the PR to acknowledge that these "quirks" are consistent with underlying markup format but might be confusing to newcomers?

README.md
liskin: You're right that this behaviour is on the YAML side of things (and is actually wrong according to the spec: http://yaml.org/spec/1.1/#id932806), but I find it very useful that it behaves as it does, and would actually propose to add tests for this behaviour and file bug reports to https://hackage.haskell.org/package/yaml if the behaviour ever changes. Here are two examples why:

Aliases with overriding:

    executables:
        strava-gear-cmdline:
            &default_executable
            main: strava-gear-cmdline.hs
            dependencies: [strava-gear]
            ghc-options:
                - -threaded

        strava-gear-web:
            <<: *default_executable
            main: strava-gear-web.hs

Predefined aliases:

    executables:
        &default_executable
        dependencies: [strava-gear]
        ghc-options:
            - -threaded

    executables:
        strava-gear-cmdline:
            <<: *default_executable
            main: strava-gear-cmdline.hs

        strava-gear-web:
            <<: *default_executable
            main: strava-gear-web.hs

The second example is a massive hack, but I'm not aware of any other way to do this given that hpack insists on printing warnings for unused fields even if I prefix them with underscores.

sol:
@deech You are welcome!  Please give it a try!

README.md
sol: @liskin Hey, interesting!  How about we don't emit warnings for unknown top-level field that starts with an underscore?  That way you can use fields that start with underscores for building kind of a `hpack` library.

I would certainly accept a patch.  So if you think thats the way to go, then feel free to open a PR!

liskin: Here you go: https://github.com/sol/hpack/pull/146 :-)

README.md
```diff
@@ -134,6 +134,49 @@ becomes
...
+
+    when:
+      - condition: os(darwin)
+        extra-lib-dirs: lib/darwin
+      - condition: os(linux)
+        extra-lib-dirs: /usr/lib/x64
+
+becomes
+
+    if os(darwin)
```
kanishka-azimi: Minor note generally, but cabal library converts Darwin to osx

sol:
@deech `hpack` now warns on duplicate fields.  So I think the main source of error has been removed!  From what I understand, what is left is still that `!` needs quoting.  I'll cherry-pick the corresponding parts from you PR when I have time to do so.

PR #142, opened 2017-01-06
Use binary I/O for cabal files

snoyberg:
This makes character encoding of UTF-8 explicit, and avoids problems
when a different system locale is set. For more information, see:
http://www.snoyman.com/blog/2016/12/beware-of-readfile

### DIFF
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -1,4 +1,4 @@
--- This file has been generated from package.yaml by hpack version 0.14.1.
+-- This file has been generated from package.yaml by hpack version 0.15.0.
 --
 -- see: https://github.com/sol/hpack
 
@@ -28,6 +28,7 @@ library
   build-depends:
       base >= 4.7 && < 5
     , base-compat >= 0.8
+    , bytestring
     , deepseq
     , directory
     , filepath
@@ -59,6 +60,7 @@ executable hpack
   build-depends:
       base >= 4.7 && < 5
     , base-compat >= 0.8
+    , bytestring
     , deepseq
     , directory
     , filepath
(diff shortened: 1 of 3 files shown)


snoyberg:
FYI, this came about when researching this build failure:

https://ci.appveyor.com/project/harendra-kumar/unicode-transforms/build/1.0.36

CC @harendra-kumar

sol:
@snoyberg thanks for the PR.  I think there is at least some one use of `readFile` somewhere too. Patches welcome.

sol:
@snoyberg A side effect of this change is that we loose newline conversion on Windows.

While I assume it's sensible that the `.cabal`-file is encoded with UTF-8**, newline conversion is usually done by Git, so we may prefer to use the systems native newline handling when writing / reading files.

** Still, right now I'm puzzled encoding `Cabal` uses when reading `.cabal` files...

sol:
> ** Still, right now I'm puzzled encoding Cabal uses when reading .cabal files...
```
package.cabal:
    a Unicode UTF-8 text file containing a package description. For details of the syntax of this file, see the section on package descriptions.
```
https://www.haskell.org/cabal/users-guide/developing-packages.html#creating-a-package

So that is fine.

tfausak:
Just chiming in to say that Cabal files are indeed UTF-8 in practice (with [one exception](https://github.com/tfausak/grawlix/blob/3189fa22614d57f284719d25b203b59c318d4465/library/Grawlix/Sync.hs#L156)). The currently unreleased Parsec-based parser will [use byte strings](https://github.com/haskell/cabal/blob/5cf994339466dec180726bac38784f311a88f97c/Cabal/Distribution/PackageDescription/Parsec.hs#L105) instead of strings. That explicitly [uses UTF-8](https://github.com/haskell/cabal/blob/5cf994339466dec180726bac38784f311a88f97c/Cabal/Distribution/Utils/Generic.hs#L167) behind the scenes.

PR #143, opened 2017-01-08
Replace textual readFile with binary readFile

snoyberg:
Follow-up to #142.

Could you ping me when this is released to Hackage? I'd like to get this updated version into Stack.


sol:
@snoyberg On Hackage, thanks a lot!

PR #144, opened 2017-01-09
Allow to "!include" yaml files

liskin:
There seems to be little reason not to allow this. Security can hardly be argued here as cabal custom setups and template haskell are far more powerful in this regard.

So anyway, here's an example of usage:

package.yaml:

```yaml
# ...

tests:
    hlint: !include "../common/hlint.yaml"
```

hlint.yaml:

```yaml
source-dirs: test
main: hlint.hs
ghc-options:
    -Wall
    -threaded
    -with-rtsopts=-N
dependencies:
    - base
    - hlint
```

Also, due to how Data.Yaml works, this can be abused to provide entire libraries of snippets:

package.yaml:

```yaml
# if "xxx:" is used instead of a known field, a warning is emitted :-(
name: !include "../common/lib.yaml"

name: example1
version: '0.1.0.0'
synopsis: Example
<<: *legal

<<: *defaults

library:
    source-dirs: src

tests:
    hlint: *test_hlint
```

lib.yaml:

```yaml
- &legal
    maintainer: Some One <someone@example.com>
    copyright: (c) 2017 Some One
    license: BSD3

- &defaults
    dependencies:
        - base
        - containers
    ghc-options:
        - -Wall
        - -Werror

- &test_hlint
    source-dirs: test
    main: hlint.hs
    dependencies: [hlint]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 ### Added
 - custom-setup section
   [#148](https://github.com/sol/hpack/pull/148)
+- Support `!include` directives
+  [#144](https://github.com/sol/hpack/pull/144)
 
 ## [0.16.0] - 2017-01-11
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+
 ## [Unreleased]
 ### Added
 - custom-setup section
+  [#148](https://github.com/sol/hpack/pull/148)
+
+## [0.16.0] - 2017-01-11
+### Added
+- Warn when `name` is missing [#109](https://github.com/sol/hpack/issues/109)
+- Support globs in `c-sources`
+  [#123](https://github.com/sol/hpack/pull/123)
+
+### Changed
+- Use binary I/O for cabal files avoiding problems with non-UTF-8 locales
+  [#142](https://github.com/sol/hpack/pull/142)
+  [#143](https://github.com/sol/hpack/pull/143)
+- Fix rendering of `.` as directory (cabal syntax issue)
+  [#125](https://github.com/sol/hpack/pull/125)
+  [#119](https://github.com/sol/hpack/issues/119)
+  [#67](https://github.com/sol/hpack/issues/67)
+
+[Unreleased]: https://github.com/sol/hpack/compare/0.16.0...HEAD
+[0.16.0]: https://github.com/sol/hpack/compare/0.15.0...0.16.0
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ These fields are merged with all library, executable, test, and benchmark compon
 | `ghc-prof-options` | · | | |
 | `cpp-options` | · | | |
 | `cc-options` | · | | |
-| `c-sources` | · | | |
+| `c-sources` | · | | Accepts [glob patterns](#file-globbing) |
 | `extra-lib-dirs` | · | | |
 | `extra-libraries` | · | | |
 | `include-dirs` | · | | |
--- a/README.md
+++ b/README.md
@@ -190,6 +190,10 @@ It is possible to use YAML [anchors][yaml-anchor] (`&`), [aliases][yaml-alias]
 (`*`) and [merge keys][yaml-merge] (`<<`) to define fields and reference them
 later.
 
+[yaml-anchor]: http://yaml.org/spec/1.1/#anchor/syntax
+[yaml-alias]: http://yaml.org/spec/1.1/#alias/syntax
+[yaml-merge]: http://yaml.org/type/merge.html
+
 ```yaml
 executables:
   my-exe-1: &my-exe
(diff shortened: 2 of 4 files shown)


sol:
In general, I'm open to this change.  But I would like to get some more feedback from `hpack` users first, just to be sure that there are no concerns.

`stack` includes `hpack` @mgsloan @snoyberg 

Also @soenkehahn any comments?

sol:
@liskin Also, regarding the build failure, apparently the `yaml` library has different show instances depending on GHC version now?  We might need CPP to fix this.  Can you look into that too?

liskin:
The build failure seems to be this: https://github.com/haskell/directory/issues/44#issuecomment-186575172
I have no idea which version of `directory` tinc brings in scope but it's likely the buggy one. Shall I go the `#if MIN_VERSION_directory(1, 2, 3)` route or can we just upgrade `directory` in Travis builds?

sol:
@liskin `tinc` tries to use the `directory` version that comes with GHC if possible.

sol:
@liskin if you can address the following points, I'm fine with merging this:

 - dig into the build failure
 - add documentation
 - as per #149 add a changelog entry

liskin:
I pushed a workaround for the failing tests, but it's (unsuprisingly) a bit ugly. Maybe I should submit it to the yaml package instead. @snoyberg what do you think?

snoyberg:
I think it makes sense for yaml.

liskin:
Okay, I'll prepare a pull req there later this evening. :-)

liskin:
So, here you go: https://github.com/snoyberg/yaml/pull/104

@sol But I guess we still want the workaround here as well, right? Anyway, I'll write the docs and changelog tomorrow.

sol:
@liskin If we can fix this in `yaml` we don't need a workaround here anymore.  I'm fine if the tests pass with the latest version of `yaml`

I'm willing to accept that the exception behavior will be different with older versions of `yaml` for the sake of code simplicity.

liskin:
Okay, yaml 0.8.21.2 is on hackage so I dropped the ugly commit, added documentation, enhanced the changelog and added an entry for this.

liskin:
Finally, green build checks! :-)

test/Hpack/ConfigSpec.hs
```diff
@@ -1151,3 +1151,25 @@ spec = do
...
+      let included = "authors.yaml"
+      withPackageConfig [i|
+        author: !include "#{included}"
+        |]
+        (writeFile included [i|
+          - A1
+          - A2
+        |]
+        )
+        (packageAuthor >>> (`shouldBe` ["A1", "A2"]))
```
sol: This two things are united tested in the `yaml` library.  I vote for removing them here.

src/Hpack/Yaml.hs
```diff
@@ -1,7 +1,8 @@
 {-# LANGUAGE RecordWildCards #-}
 module Hpack.Yaml where
 
-import           Data.Yaml
+import           Data.Yaml hiding (decodeFile, decodeFileEither)
+import           Data.Yaml.Include
```
sol: :+1:

README.md
```diff
@@ -214,9 +218,63 @@ executables:
...
+    - containers
+  ghc-options:
+    - -Wall
+    - -Werror
+
+- &test_hlint
+  source-dirs: test
+  main: hlint.hs
+  dependencies: [hlint]
+```
```
sol: :+1:

test/Hpack/ConfigSpec.hs
liskin: Oh, I considered it part of "hpack specification" and naturally wrote a hspec test for that. :-)

CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Haskell Package Versioning Policy](https://pvp.haskell.org/).
```
sol: I agree with the PVP for my code.  But the current wording of the PVP also contains this wording:

> Note that modifying imports or depending on a newer version of another package may cause extra orphan instances to be exported and thus force a major version change.

Some people interpret this as: In theory one of your dependencies (e.g. `directory`) could add an orphan instance, for that reason following the PVP means you have to specify upper bounds for all your dependencies.  I'm not going to do that.

CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
...
+- Use binary I/O for cabal files avoiding problems with non-UTF-8 locales
+  [#142](https://github.com/sol/hpack/pull/142)
+  [#143](https://github.com/sol/hpack/pull/143)
+- Fix rendering of `.` as directory (cabal syntax issue)
+  [#125](https://github.com/sol/hpack/pull/125)
+  [#119](https://github.com/sol/hpack/issues/119)
+  [#67](https://github.com/sol/hpack/issues/67)
+
+[Unreleased]: https://github.com/sol/hpack/compare/0.16.0...HEAD
+[0.16.0]: https://github.com/sol/hpack/compare/0.15.0...0.16.0
```
sol: I'm on the fence with the hyperlinks in the change log.  I agree that it looks nice, but I may not have the bandwidth to keep up with doing it like this in the future.  Let's keep it that way for now and see.

CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
...
 ### Added
 - custom-setup section
+  [#148](https://github.com/sol/hpack/pull/148)
+- Support `!include` directives
+  [#144](https://github.com/sol/hpack/pull/144)
+
+## [0.16.0] - 2017-01-11
+### Added
+- Warn when `name` is missing [#109](https://github.com/sol/hpack/issues/109)
+- Support globs in `c-sources`
```
sol: I think this change never made it into the README.  Extra points if you can fix that ;)

sol: Basically we want this "Accepts glob patterns" comments, same to what we have for other fields

CHANGELOG.md
sol: Let's just remove this paragraph.

sol:
@liskin Hey, thanks for taking care of this and all the other issues that spawned from it.  I think this is a pretty awesome change.

If you can address my bikeshedding, it's good for merge.

test/Hpack/ConfigSpec.hs
sol: I don't fully agree with JBrain, but as a reference: http://blog.thecodewhisperer.com/permalink/integrated-tests-are-a-scam

liskin: "Interesting" article. Too bad it offers no alternatives. Not having integration tests at all and just having it blow up in production is, in my humble opinion, not a better alternative. :-)

But I can agree that slow tests are really bad, but I think that's a technical problem, not a fundamental one. For a long time I wanted qemu to support superfast copy on write snapshots via forking (with readonly file/disk-backed storage, obviously) which I think could help a lot, but never had enough time/budget to implement that.

Anyway, I dropped the two tests.

liskin:
And now the Windows build blows up when building happy. That can't be my fault, can it?

liskin:
Pushed again, everything's green again.

sol:
@liskin Thanks a lot!

In general, my bandwidth is limited for medical reasons.  I try to keep up with important stuff, but I'm grateful for any help.  If you want to further help with things, then go ahead and bump the version.  Also what is your Hackage account name?

liskin:
https://github.com/sol/hpack/pull/154

I'm `TomasJanousek` on Hackage.

PR #146, opened 2017-01-12
Silence warnings for unknown top-level fields starting with an underscore

liskin:
This lets users declare standalone YAML anchors and reuse them using YAML
aliases later:

```yaml
_: &mydeps
 - base
 - containers
```

```yaml
_my_library_of_hpack_stuff:
 - &mydeps [base, containers]
 - &myflags [-threaded, -Wall]
```

Also, this pull request adds a test that field overriding works as
expected and extends the README with a section on not repeating
yourself using these YAML tricks.

See also #144, which is a followup (logically, but not chronologically)
to this and enables putting those _libraries_of_hpack_stuff to separate
files and including them in multiple package.yamls across a larger
codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,40 @@ Glob patterns are expanded according to the following rules:
    separators)
  - `?`, `*` and `**` do not match a `.` at the beginning of a file/directory
 
+### Not repeating yourself
+
+It is possible to use YAML [anchors][yaml-anchor] (`&`), [aliases][yaml-alias]
+(`*`) and [merge keys][yaml-merge] (`<<`) to define fields and reference them
+later.
+
+```yaml
+executables:
+  my-exe-1: &my-exe
+    main: my-exe-1.hs
+    dependencies: [base, my-lib]
+    ghc-options: [-threaded]
+  my-exe-2:
+    <<: *my-exe
+    main: my-exe-2.hs
+```
+
+Warnings for unknown fields will not be emitted for top-level fields starting
+with an underscore, so you can declare global aliases too:
+
+```yaml
+_exe-ghc-options: &exe-ghc-options
+  - -threaded
+  - -rtsopts
+
+executables:
+  my-exe-1:
+    ghc-options: *exe-ghc-options
+```
+
+[yaml-anchor]: http://yaml.org/spec/1.1/#anchor/syntax
+[yaml-alias]: http://yaml.org/spec/1.1/#alias/syntax
+[yaml-merge]: http://yaml.org/type/merge.html
+
 ### Slides
 
  - Slides from my talk about `hpack` at the Singapore Haskell meetup:
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -48,7 +48,7 @@ import           Data.Data
 import           Data.Map.Lazy (Map)
 import qualified Data.Map.Lazy as Map
 import qualified Data.HashMap.Lazy as HashMap
-import           Data.List.Compat (nub, (\\), sortBy)
+import           Data.List.Compat (nub, (\\), sortBy, isPrefixOf)
 import           Data.Maybe
 import           Data.Ord
 import           Data.String
@@ -128,6 +128,9 @@ class HasFieldNames a where
   default fieldNames :: (HasTypeName a, Selectors (Rep a)) => Proxy a -> [String]
   fieldNames proxy = map (hyphenize $ typeName proxy) (selectors proxy)
 
+  ignoreUnderscoredUnknownFields :: Proxy a -> Bool
+  ignoreUnderscoredUnknownFields _ = False
+
 data CaptureUnknownFields a = CaptureUnknownFields {
   captureUnknownFieldsFields :: [FieldName]
 , captureUnknownFieldsValue :: a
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -177,6 +177,40 @@ Glob patterns are expanded according to the following rules:
...
+  my-exe-2:
+    <<: *my-exe
+    main: my-exe-2.hs
+```
+
+Warnings for unknown fields will not be emitted for top-level fields starting
+with an underscore, so you can declare global aliases too:
+
+```yaml
+_: &exe-ghc-options
```
sol: Let's use a variation of the common convention where the field name and anchor name are the same:

```yaml
_exe-ghc-options: &exe-ghc-options
  - ...
```

test/Hpack/ConfigSpec.hs
```diff
@@ -642,6 +645,13 @@ spec = do
         }
         )
 
+    it "allows overriding fields and takes the last value" $ do
+      withPackageConfig_ [i|
+        name: n1
+        name: n2
+        |]
+        (packageName >>> (`shouldBe` "n2"))
```
sol: Now that we ignore warnings for fields that start with an underscore, is this still critical?  I think it's better not to rely on that behavior.  Please remove this test if you agree.

src/Hpack/Config.hs
```diff
@@ -150,11 +153,14 @@ instance FromJSON (CaptureUnknownFields FlagSection) where
...
 getUnknownFields v _ = case v of
-  Object o -> unknown
+  Object o -> ignoreUnderscored unknown
     where
       unknown = keys \\ fields
       keys = map T.unpack (HashMap.keys o)
       fields = fieldNames (Proxy :: Proxy a)
+      ignoreUnderscored
+        | warnUnderscoredUnknownFields (Proxy :: Proxy a) = id
+        | otherwise = filter (not . isPrefixOf "_")
```
sol: I like your code, it's quite elegant and extensible if we ever want to change the behavior in the future.  Yet, being advocate of TDD, I usually tend to go with the simplest solution that solves the problem and defer more clever solutions until we actually need them.

Further down we have a line

```haskell
mkPackage dir (CaptureUnknownFields unknownFields globalOptions@Section{sectionData = PackageConfig{..}}) = do
```

Here you could make the changes with a two liner (using view patterns):

```haskell
ignoreUnderscored = filter (not . isPrefixOf "_")

mkPackage dir (CaptureUnknownFields (ignoreUnderscored -> unknownFields) globalOptions@Section{sectionData = PackageConfig{..}}) = do
```

test/Hpack/ConfigSpec.hs
```diff
@@ -323,9 +324,11 @@ spec = do
           - condition: impl(ghc)
             bar: 23
             baz: 42
+            _qux: 66
         |]
         (`shouldBe` [
-          "Ignoring unknown field \"bar\" in package description"
+          "Ignoring unknown field \"_qux\" in package description"
```
sol: If you go with the two-liner solution this test case becomes redundant.

README.md
sol: My main motivation here is that if the user uses `_` multiple times we rely on the non-standard `libyaml` behavior.  This is why I prefer an example that avoids this problem by including the reference name in the field name.

test/Hpack/ConfigSpec.hs
liskin: It is still critical because otherwise merge and override (https://github.com/sol/hpack/pull/146/commits/d10a66db70aff2c225622fcdda929d15d92f7312#diff-04c6e90faac2675aa89e2176d2eec7d8R194) wouldn't work.

src/Hpack/Config.hs
liskin: Oh, I had almost the exact same code but then I realized we agreed on top-level unknown fields and rewrote it this way. :-(
But sure, I can simplify this.

test/Hpack/ConfigSpec.hs
sol: My assumption was that this will always work and is not violating the spec.  But then again, I'm not sure.

src/Hpack/Config.hs
sol: I think the code I suggested will behave exactly the same way as your code (that is only ignore top-level fields that start with an underscore).

test/Hpack/ConfigSpec.hs
liskin: I'm not really sure how to interpret the spec regarding this, but I think that testing for the merge&override case specifically instead of this double definition would be a good idea anyway. :-)

src/Hpack/Config.hs
liskin: It won't, it will ignore this: https://github.com/sol/hpack/pull/146/files/d10a66db70aff2c225622fcdda929d15d92f7312#diff-af92d09f3a5eb3bf28df295b3a1fa13fR327 as well and I didn't consider that a top-level field. But I really don't have any strong feelings about this. If you think the simplification is worth it, let's do it that way. :-)

test/Hpack/ConfigSpec.hs
sol: Yes, I'm ok with that.

src/Hpack/Config.hs
sol: Ah, I see, it's because we treat unknown fields from top-level conditionals the same way as unknown top-level fields.  In that case let's go with your solution.

Given that there is only one more thing I would ask.  Can we rename `warnUnderscoredUnknownFields` to `ignoreUnderscoredUnknownFields` inverse the bools.  That code is not really concerned with warnings, but rather with collecting unknown field names.

test/Hpack/ConfigSpec.hs
sol: As discussed we actually want to keep this test case.

liskin:
I pushed an updated version that hopefully addresses everything we agreed upon.

Issue #147, opened 2017-01-16
Support custom-setup stanza


mitchellwrosen:
Currently the only field in a `custom-setup` stanza is `setup-depends`, so maybe `setup-depends` could be added as a new top-level `hpack` field.

PR #148, opened 2017-01-16
Add setup-depends

mitchellwrosen:
Addresses #147 

I did this rather quickly, please let me know if you see any problems.

Couple notes...

1. `setup-depends` is the only field inside a `custom-setup` stanza (currently), so I just made it a top-level field. Should it go inside a `custom-setup` stanza a la cabal instead?
2. I'm not sure if having a `setup-depends` implies `build-type: Custom` or not.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -24,6 +24,7 @@ module Hpack.Config (
 , GitUrl
 , GitRef
 , GhcOption
+, CustomSetup(..)
 , Section(..)
 , Library(..)
 , Executable(..)
@@ -65,7 +66,32 @@ import           Hpack.Util
 import           Hpack.Yaml
 
 package :: String -> String -> Package
-package name version = Package name version Nothing Nothing Nothing Nothing Nothing Nothing [] [] [] Simple Nothing Nothing Nothing [] [] [] Nothing Nothing [] [] []
+package name version = Package {
+    packageName = name
+  , packageVersion = version
+  , packageSynopsis = Nothing
+  , packageDescription = Nothing
+  , packageHomepage = Nothing
+  , packageBugReports = Nothing
+  , packageCategory = Nothing
+  , packageStability = Nothing
+  , packageAuthor = []
+  , packageMaintainer = []
+  , packageCopyright = []
+  , packageBuildType = Simple
+  , packageLicense = Nothing
+  , packageLicenseFile = Nothing
+  , packageTestedWith = Nothing
+  , packageFlags = []
+  , packageExtraSourceFiles = []
+  , packageDataFiles = []
+  , packageSourceRepository = Nothing
+  , packageCustomSetup = Nothing
+  , packageLibrary = Nothing
+  , packageExecutables = []
+  , packageTests = []
+  , packageBenchmarks = []
+  }
 
 renamePackage :: String -> Package -> Package
 renamePackage name p@Package{..} = p {
@@ -148,6 +174,9 @@ instance (HasFieldNames a, FromJSON a) => FromJSON (CaptureUnknownFields (Sectio
     where
       unknownSectionFields = getUnknownFields v (Proxy :: Proxy (Section a))
 
+instance FromJSON (CaptureUnknownFields CustomSetupSection) where
+  parseJSON = captureUnknownFields
+
 instance FromJSON (CaptureUnknownFields FlagSection) where
   parseJSON = captureUnknownFields
 
@@ -162,6 +191,15 @@ getUnknownFields v _ = case v of
         | ignoreUnderscoredUnknownFields (Proxy :: Proxy a) = filter (not . isPrefixOf "_")
         | otherwise = id
   _ -> []
+
+data CustomSetupSection = CustomSetupSection {
+  customSetupSectionDependencies :: Maybe (List Dependency)
+} deriving (Eq, Show, Generic)
+
+instance HasFieldNames CustomSetupSection
+
+instance FromJSON CustomSetupSection where
+  parseJSON = genericParseJSON_
 
 data LibrarySection = LibrarySection {
   librarySectionExposed :: Maybe Bool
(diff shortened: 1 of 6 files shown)


src/Hpack/Config.hs
```diff
@@ -65,7 +65,32 @@ import           Hpack.Util
...
+  , packageFlags = []
+  , packageExtraSourceFiles = []
+  , packageDataFiles = []
+  , packageSourceRepository = Nothing
+  , packageSetupDepends = []
+  , packageLibrary = Nothing
+  , packageExecutables = []
+  , packageTests = []
+  , packageBenchmarks = []
+  }
```
sol: This was about time.  Much better now :+1:

test/Hpack/ConfigSpec.hs
```diff
@@ -642,6 +642,14 @@ spec = do
         }
         )
 
+    it "accepts setup-depends" $ do
+      withPackageConfig_ [i|
+        setup-depends:
+          - foo-1.0
+          - bar-2.0
+        |]
+        (packageSetupDepends >>> (`shouldBe` ["foo-1.0", "bar-2.0"]))
```
sol: I would assume this has to follow the same format as other dependencies, say `foo == 1.0` instead of `foo-1.0`?

test/Hpack/RunSpec.hs
```diff
@@ -103,6 +103,19 @@ spec = do
         , "  default-language: Haskell2010"
         ]
 
+    it "includes setup-depends" $ do
+      renderPackage_ package {packageSetupDepends = ["foo-1.0", "bar-2.0"]} `shouldBe` unlines [
```
sol: Same here.

test/Hpack/ConfigSpec.hs
mitchellwrosen: Whoops... yup! Moment.

mitchellwrosen: fixed

sol:
@mitchellwrosen After thinking about it for a quite a bit, I would go for the following concrete syntax:

```yaml
custom-setup:
  dependencies:
    - foo
    - bar
```

Mainly for consistency, as we already use `dependencies` in `hpack` for other dependencies.

@mitchellwrosen what do you think?  If you are fine with this, please go ahead and make the changes.

I'm also tempted to imply `build-type: Custom` if there is a `custom-setup` field.

mitchellwrosen:
@sol Sounds good, I like both of those changes. Pushed a new patch.

src/Hpack/Config.hs
```diff
@@ -463,12 +507,27 @@ data SourceRepository = SourceRepository {
...
+    name :: String
     (nameWarnings, name) = maybe (["Package name not specified, inferred " ++ show inferredName], inferredName) ((,) []) packageConfigName
       where inferredName = takeBaseName dir
 
+  let
+    customSetupResult :: Maybe ([String], CustomSetup)
+    customSetupResult = toCustomSetup packageConfigBuildType <$> mCustomSetupSection
+
+    customSetupWarnings :: [String]
+    customSetupWarnings = fromMaybe [] (fmap fst customSetupResult)
```
sol: Can this be simplified to

```haskell
customSetupWarnings = maybe [] fst customSetupResult
```
?

src/Hpack/Config.hs
```diff
@@ -490,7 +549,13 @@ mkPackage dir (CaptureUnknownFields unknownFields globalOptions@Section{sectionD
   (dataFilesWarnings, dataFiles) <-
     expandGlobs "data-files" dir (fromMaybeList packageConfigDataFiles)
 
-  let pkg = Package {
+  let defaultBuildType :: BuildType
+      defaultBuildType =
+        if isJust mCustomSetup
+          then Custom
+          else Simple
```
sol:
```haskell
defaultBuildType = maybe Simple (const Custom) mCustomSetup
```

src/Hpack/Config.hs
```diff
@@ -603,6 +674,20 @@ expandCSources dir sect@Section{..} = do
...
 
+toCustomSetup :: Maybe BuildType -> CustomSetupSection -> ([String], CustomSetup)
+toCustomSetup mBuildType CustomSetupSection{..} = (warnings, setup)
+  where
+    warnings :: [String]
+    warnings =
+      case mBuildType of
+        Nothing -> []
+        Just Custom -> []
+        Just buildType -> ["Existence of `custom-setup` stanza implies `build-type: Custom`, but package.yaml sets `build-type: " ++ show buildType ++ "`"]
```
sol: I wouldn't even care about this and leave it to `cabal` the case when a user specifies nonsense here.

Are you comfortable with removing this warning?

src/Hpack/Run.hs
```diff
@@ -68,8 +68,13 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
     dataFiles :: Element
     dataFiles = Field "data-files" (LineSeparatedList packageDataFiles)
 
+    sourceRepository :: [Element]
     sourceRepository = maybe [] (return . renderSourceRepository) packageSourceRepository
 
+    setupDepends :: [Element]
+    setupDepends = maybe [] (return . renderCustomSetup) packageCustomSetup
```
sol: On the nitpicking side of things: This should be `customSetup` now, not `setupDepends`

test/Hpack/ConfigSpec.hs
```diff
@@ -642,6 +642,58 @@ spec = do
...
+        withPackageWarnings_ [i|
+          name: foo
+          build-type: Simple
+          custom-setup:
+            dependencies:
+              - base
+          |]
+          (`shouldBe` [
+            "Existence of `custom-setup` stanza implies `build-type: Custom`, but package.yaml sets `build-type: Simple`"
+          ])
```
sol: As stated above, I would remove this test case.

sol:
@mitchellwrosen Apart from my comments, this looks really good now.  Please address and squash into a single commit.

Please also feel free to merge / push to master in case there is nothing to discuss anymore.

sol:
As per #149, please also add a changelog entry

sol:
Also needs to be rebased now, sorry for that!

src/Hpack/Config.hs
mitchellwrosen: :+1:

src/Hpack/Run.hs
mitchellwrosen: Whoops, forgot to rename

mitchellwrosen:
Oops... fixed up and pushed but I forgot the changelog entry. Should I make a new changelog.md file? I don't see one.

mitchellwrosen:
Ah, nevermind, I see that the issue's new. Ok, added a changelog entry :+1:

sol:
Thanks a lot!

PR #151, opened 2017-01-24
Replace identically-named fields with a dot in README

mitchellwrosen:
This makes it more clear at a glance where hpack has diverged from cabal


sol:
Fine with me.

PR #152, opened 2017-01-27
Add motivation section to README.md

### LINKED ISSUE #140: Suggestion: High-level summary in cabal file and readme
Just a suggestion, feel free to ignore. I don't directly use hpack but I depend on a few packages that do so I have to ensure that it runs. Every time I get curious I look it up on hackage and then on github and all I see is a description "An alternative format for Haskell packages" which doesn't really convey any meaningful info to me and then an examples and file format documentation (both of which look great). Might I suggest a brief blurb on both the cabal description and readme that describes the motivating case where you'd use hpack or what problems it solves?

MichaelXavier:
This resolves #140. The list may be phrased more elegantly and added
to, but its better to have at least something there ASAP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,25 @@
 # hpack: An alternative format for Haskell packages
 
+## Motivation
+
+hpack makes the following improvements over the standard cabal file
+format:
+
+* YAML syntax is more ubiquitous than cabal syntax and a bit easier to
+  write. It is also very easy to parse from other languages.
+* `exposed-modules`/`other-modules` are set automatically. In a
+  standard cabal file these are easy to forget when new modules are
+  added and will result in obscure linker errors or worse.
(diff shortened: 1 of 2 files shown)


sol:
I resolved the conflicts and pushed to `master`.

sol:
I originally planed to improve on top of this, but I don't have the spare cycles to carve this out properly.  Instead I ended up reverting and pushing this instead https://github.com/sol/hpack/commit/ac336041732d0e1416118c34defe43647337f42e.

That said, if somebody wants to put more effort into this, that would be still welcome.

MichaelXavier:
Is there something inaccurate with the bullet points I added? Seems like it had the same content as the slide deck if not more except for it was inline with the page the user would be visiting rather than spread across 18 slides. This kind of feels like a step back to me.

PR #156, opened 2017-01-31
Set cabal-version >= 1.24 if custom-setup exists

mitchellwrosen:


sol:
As I understand it, `custom-setup` is ignored by older versions of Cabal. So having this should be backward compatible.

I understand it as progressive enhancement: If you have a recent version of `cabal-install` then it will take care of installing the dependencies for you.

If your version of `cabal-install` is too old, you have to install the dependencies yourself.  My initial assumption is that installing the dependencies yourself is still better than not being able to build the package at all. In some case the dependency may even be already in place (say if the dependency comes with GHC, like `directory).

Given this reasoning, do you still think this is useful?

mitchellwrosen:
That does make sense, although currently hpack doesn't let you set `cabal-version` at all. So, if for some reason you wanted to require cabal 1.24+, maybe to prevent people writing bug reports about missing dependencies or something, you wouldn't be able to.

But, for now I agree allowing a lower bound is preferable. I actually didn't know cabal ignored unknown fields!

PR #164, opened 2017-03-25
Add shortcut executable section

### LINKED ISSUE #161: Shortcuts for executables and tests
I think it would be nice to have shortcuts for defining executables and tests (and probably benchmarks too). Pretty much every package I define with executables and tests looks like this:

``` yaml
name: some-package
library: ...
executables:
  some-package: ...
tests:
  test: ...
```

That is, I have a single executable with the same name as the package and a single test suite with the name "test". I always wish it could be shortened to this:

``` yaml
name: some-package
library: ...
executable: ...
test: ...
```

The top level `executable: x` would be transformed into `executables: { some-package: x }`, and the `test` part would do the same. 

Is there any interest in this feature?

tfausak:
This fixes #161. It adds a new `executable` section that can be used in addition to or alongside the old `executables` section. This new section allows you to simplify executable definition from this:

``` yaml
name: example
executables:
  example: { ... }
```

To this:

``` yaml
name: example
executable: { ... }
```

That is, it defines a new executable and uses the package name as the name of the executable. 

If you define an `executable` section, you can still define an `executables` section (and vice versa). For example:

``` yaml
name: example
executable: { ... }
executables:
  example-extra: { ... }
```

If you define both sections and one of the `executables` has the same name as the package, it will be silently discarded in favor of the `executable` section. For instance:

``` yaml
name: example
executable: { ... } # this one wins
executables:
  example: { ... } # this one loses
```

This is a little surprising. I would prefer to make it either a warning or an error, but I wanted to make sure I'm barking up the right tree with this pull request before I spend more time on it 😄

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -523,16 +523,15 @@ mkPackage dir (CaptureUnknownFields unknownFields globalOptions@Section{sectionD
   let
     executableWarnings :: [String]
     executableSections :: [(String, CaptureUnknownFields (Section ExecutableSection))]
-    (executableWarnings, executableSections) = case packageConfigExecutable of
-      Nothing -> ([], toList packageConfigExecutables)
-      Just executable -> case packageConfigExecutables of
-        Nothing -> ([], [(packageName_, executable)])
-        Just executables -> case Map.insertLookupWithKey (\_ x _ -> x) packageName_ executable executables of
-          (Nothing, m) -> ([], Map.toList m)
-          (Just _, m) ->
-            ( ["Ignoring field " ++ show packageName_ ++ " in executables section in favor of implicit \"executable\" section"]
-            , Map.toList m
-            )
+    (executableWarnings, executableSections) =
+      case (packageConfigExecutable, packageConfigExecutables) of
+        (Nothing, Nothing) -> ([], [])
+        (Just executable, Nothing) -> ([], [(packageName_, executable)])
+        (Nothing, Just executables) -> ([], Map.toList executables)
+        (Just _, Just executables) ->
+          ( ["Ignoring executable section because executables section exists"]
+          , Map.toList executables
+          )
 
     mLibrary :: Maybe (Section Library)
     mLibrary = fmap snd libraryResult
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -599,8 +599,8 @@ mkPackage dir (CaptureUnknownFields unknownFields globalOptions@Section{sectionD
         ++ flagWarnings
         ++ maybe [] (formatUnknownFields "custom-setup section") (captureUnknownFieldsFields <$> packageConfigCustomSetup)
         ++ maybe [] (formatUnknownFields "library section") (captureUnknownFieldsFields <$> packageConfigLibrary)
-        ++ formatUnknownSectionFields "executable" executableSections
-        ++ formatUnknownSectionFields "test" testsSections
+        ++ formatUnknownSectionFields (isJust packageConfigExecutables) "executable" executableSections
+        ++ formatUnknownSectionFields True "test" testsSections
         ++ formatMissingSourceDirs missingSourceDirs
         ++ libraryWarnings
         ++ executableWarnings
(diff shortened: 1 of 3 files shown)


sol:
In terms of code the additional complexity looks ok. What about user documentation? How would the `README` be affected?

soenkehahn:
@sol asked me for a second opinion on this, so here it is: I think this is a good idea. (And yes, I think having one section ignored in case of overlapping names is bad and should produce a warning.)

I share @sol's concern about the added complexity. I think code-wise the added complexity is worth it, but I wonder about the documentation. @tfausak: Would you be willing to create a draft for the documentation in the readme and include that in this PR? So that we can see how that would look like? (I think the documentation should clearly state that this is a shorthand for `executables`.)

So generally you're barking up the right tree, I think. :) And thanks for working on this!

tfausak:
Thanks for the feedback! I updated the README and added a warning. The warning looks like this:

``` yaml
name: some-package
executable: { ... }
executables:
  some-package: { ... }
# Ignoring field "some-package" in executables section in favor of implicit "executable" section
```

sol:
@tfausak Can you rebase on `master` please + have a look at this test case:

```haskell
  it "warns on unknown fields" $ do
    withPackageWarnings_ [i|
      name: foo
      executable:
        main: Main.hs
        bar: 42
        baz: 23
      |]
      (`shouldBe` [
        "Ignoring unknown field \"bar\" in executable section"
      , "Ignoring unknown field \"baz\" in executable section "
      ]
      )
```

tfausak:
I added a test case for that, but I see the problem you're describing. It says `in executable section "foo"` rather than just `in the executable section`. I'll have to think about how to fix that.

tfausak:
After thinking about this for a while, I think the best way forward is to forbid having both the `executable` and `executables` sections at the same time. That would make it easy to generate a good error message. Also it would make things easier for the typical case of a single executable that matches the package name while not making the unusual case of multiple executables any worse.

tfausak:
I'm happy with how this works now. If you provide both `executable` and `executables` sections you'll get a warning. The `executables` section always takes precedence; they are not merged. Warnings that come from inside one of the sections will either say `in the executable section` or `in the executable section "foo"`, where "foo" is the name of that executable. Here are some concrete examples:

``` yaml
# This is the old way of doing things. No changes to how this works.
executables:
 foo:
  main: Main.hs

# This is the new shortcut I introduced. No warnings here.
executable:
 main: Main.hs

# This will generate a warning. The `executable` section will be ignored.
# In other words there will only be a `foo2` executable. No `foo` executable.
executable:
 main: Main.hs
executables:
 foo2:
  main: Main.hs

# This will generate a warning that says:
#   Ignoring unknown field "unknown" in executable section "foo"
executables:
 foo:
  main: Main.hs
  unknown: true

# This will generate a warning that says:
#   Ignoring unknown field "unknown" in executable section
executable:
 main: Main.hs
 unknown: true
```

tfausak:
Is there anything I can do to help move this along?

src/Hpack/Config.hs
```diff
@@ -586,10 +599,11 @@ mkPackage dir (CaptureUnknownFields unknownFields globalOptions@Section{sectionD
         ++ flagWarnings
         ++ maybe [] (formatUnknownFields "custom-setup section") (captureUnknownFieldsFields <$> packageConfigCustomSetup)
         ++ maybe [] (formatUnknownFields "library section") (captureUnknownFieldsFields <$> packageConfigLibrary)
-        ++ formatUnknownSectionFields "executable" executableSections
-        ++ formatUnknownSectionFields "test" testsSections
+        ++ formatUnknownSectionFields (isJust packageConfigExecutables) "executable" executableSections
+        ++ formatUnknownSectionFields True "test" testsSections
```
sol: Damn, seems we never printed warnings for unknown fields in benchmark sections.  Let me fix that first.

sol: Done in https://github.com/sol/hpack/commit/a6a13f79a5d5b84470dd29913dc757c9541af690.

sol:
@tfausak I think this is much better now! It makes a lot of sense to only allow either `executable` or `executables`, but never both at the same time.

I'm finally getting excited about this!  I would even want to take it one level further:  At some places we accept both values and list of values for the same field (e.g. `dependencies` can be one dependency or a list of dependencies).  In a similar way we could accept both an executable or a mapping from executable names to executables for `executable`.  Both of the following should be accepted:

```yaml
executable:
  main: Main.hs
```
or
```yaml
executable:
  foo:
    main: Foo.hs
  bar:
    main: Bar.hs
```
What I like about this is that we make invalid combinations (both `executable` and `executables` present unrepresentable).  I imagine it will also make the code even simpler.

sol:
@tfausak I don't expect anybody to be as excited about this as approach as I am, but is this at least something that works for you?

sol:
@tfausak I'm taking your code as a basis and giving this a shot.

tfausak:
:+1:

I didn't chase down that path because it seemed like it would be ambiguous in some cases.

sol:
> I didn't chase down that path because it seemed like it would be ambiguous in some cases.

I had that though at first, but actually I don't think it is.

sol:
@tfausak btw, make sure to use GHC 7.10.3 for development, otherwise source locations for failing tests are screwed (see https://ghc.haskell.org/trac/ghc/ticket/13285).

sol:
Proof of concept: https://github.com/sol/hpack/commit/b878242152fbb7d3d9c507472f7e49485573ef19

PR #175, opened 2017-05-18
Propagate section-specific data in conditionals.

### LINKED ISSUE #116: How to add `exposed-modules` depending on a cabal flag?
I can't manage to create a `package.yaml` that allows to add `exposed-modules` depending on a cabal flag. This is roughly what I tried:

``` yaml
  when:
  - condition: flag(with-servant-aeson-specs)
    dependencies:
    - servant-aeson-specs ==0.2.* || ==0.3.* || ==0.4.*
    exposed-modules:
    - Servant.MatrixParam.AesonSpecs
```

But I get the following message from `hpack`:

```
WARNING: Ignoring unknown field "exposed-modules" in library section
servant-matrix-param.cabal is up-to-date
```

And the cabal file is missing the `exposed-modules`.

Is there a way to do this with `hpack`?

(Here's the relevant project, if that helps: https://github.com/plow-technologies/servant-matrix-param)

scott-fleischman:
This is a working approach to allowing section-specific properties in conditionals.

The main goal is to allow library-specific properties in the body of conditionals. For example:
```
library:
  source-dirs: src
  exposed-modules: Foo
  other-modules: []
  when:
    condition: os(windows)
    exposed-modules:
    - Bar
```

Implementation details:

Effectively I propagated the type parameter for a `Section` throughout all of the conditional-related code.

This changes the existing behavior in a few instances.
* You can now specific properties like `github` within a conditional for the package (top-level properties). One previous test assumed that was invalid, and I changed the test to allow that case.
* Package-specific properties within conditionals allow underscore prefixes, because all package-specific do. It is less work to leave it like this. Note that it does require a change to a test which assumed that wasn't allowed before.
* `executableSectionMain` in `ExecutableSection` now needs to be `Maybe FilePath` instead of `FilePath`. This is so that you don't have to specify `main` in every conditional in an executable section (which would break existing code). Note that this pushes the validation that `main` is supplied to a different area in the codebase. The new validation code works but could be improved.

I added a `FlatThen` as an analogous type to `ThenElse` which handles parsing the flat `condition` property with its sibling section properties.

Note of caution: it is recommended to always set `other-modules` (and `exposed-modules` in the `library` section) at the root level when using one or both properties within a conditional. Else the recursive file selection behavior will likely conflict with what the conditional is trying to achieve. Note that in the code, the conditionals do not expand empty `exposed-modules` and `other-modules`. That is only done at the first level of the `library` and executable sections.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -8,7 +8,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 module Hpack.Config (
@@ -101,14 +101,14 @@ renamePackage name p@Package{..} = p {
   , packageBenchmarks = map (renameDependencies packageName name) packageBenchmarks
   }
 
-renameDependencies :: String -> String -> Section a -> Section a
+renameDependencies :: String -> String -> Section a b -> Section a b
 renameDependencies old new sect@Section{..} = sect {sectionDependencies = map rename sectionDependencies, sectionConditionals = map renameConditional sectionConditionals}
   where
     rename dep
       | dependencyName dep == old = dep {dependencyName = new}
       | otherwise = dep
 
-    renameConditional :: Conditional -> Conditional
+    renameConditional :: Conditional b -> Conditional b
     renameConditional (Conditional condition then_ else_) = Conditional condition (renameDependencies old new then_) (renameDependencies old new <$> else_)
 
 packageDependencies :: Package -> [Dependency]
@@ -118,7 +118,7 @@ packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . d
   ++ (concatMap sectionDependencies packageBenchmarks)
   ++ maybe [] sectionDependencies packageLibrary
 
-section :: a -> Section a
+section :: a -> Section a a
 section a = Section a [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] []
 
 packageConfig :: FilePath
@@ -167,12 +167,12 @@ captureUnknownFields v = CaptureUnknownFields unknown <$> parseJSON v
   where
     unknown = getUnknownFields v (Proxy :: Proxy a)
 
-instance (HasFieldNames a, FromJSON a) => FromJSON (CaptureUnknownFields (Section a)) where
+instance (HasFieldNames a, FromJSON a, HasFieldNames b, FromJSON b) => FromJSON (CaptureUnknownFields (Section a b)) where
   parseJSON v = do
     (unknownFields, sect) <- toSection <$> parseJSON v <*> parseJSON v
     return (CaptureUnknownFields (unknownSectionFields ++ unknownFields) sect)
     where
-      unknownSectionFields = getUnknownFields v (Proxy :: Proxy (Section a))
+      unknownSectionFields = getUnknownFields v (Proxy :: Proxy (Section a b))
 
 instance FromJSON (CaptureUnknownFields CustomSetupSection) where
   parseJSON = captureUnknownFields
@@ -214,7 +214,7 @@ instance FromJSON LibrarySection where
   parseJSON = genericParseJSON_
 
 data ExecutableSection = ExecutableSection {
-  executableSectionMain :: FilePath
+  executableSectionMain :: Maybe FilePath
 , executableSectionOtherModules :: Maybe (List String)
 } deriving (Eq, Show, Generic)
 
@@ -250,7 +250,7 @@ instance HasFieldNames a => HasFieldNames (CommonOptions a)
 instance (FromJSON a, HasFieldNames a) => FromJSON (CommonOptions a) where
   parseJSON = genericParseJSON_
 
-data ConditionalSection a = ThenElseConditional (CaptureUnknownFields (ThenElse a)) | FlatConditional (CaptureUnknownFields (Section Condition))
+data ConditionalSection a = ThenElseConditional (CaptureUnknownFields (ThenElse a)) | FlatConditional (CaptureUnknownFields (FlatThen a))
   deriving (Eq, Show)
 
 instance (FromJSON a, HasFieldNames a) => FromJSON (ConditionalSection a) where
@@ -271,10 +271,28 @@ instance FromJSON Condition where
 
 instance HasFieldNames Condition
 
+data FlatThen a = FlatThen {
+  _flatThenCondition :: String
+, _flatThenBody :: CaptureUnknownFields (Section a a)
+} deriving (Eq, Show, Generic)
+
+instance (HasFieldNames a, FromJSON a) => FromJSON (CaptureUnknownFields (FlatThen a)) where
+  parseJSON v = do
+    Condition condition <- parseJSON v
+    CaptureUnknownFields fields body <- parseJSON v
+    let unknown = fields \\ fieldNames (Proxy :: Proxy Condition)
+    return $ CaptureUnknownFields unknown (FlatThen condition (CaptureUnknownFields [] body))
+
+instance HasFieldNames a => HasFieldNames (FlatThen a) where
+  fieldNames _ = fieldNames (Proxy :: Proxy Condition) ++ fieldNames (Proxy :: Proxy a)
+
+instance (HasFieldNames a, FromJSON a) => FromJSON (FlatThen a) where
+  parseJSON v = FlatThen <$> fmap conditionCondition (parseJSON v) <*> parseJSON v
+
 data ThenElse a = ThenElse {
   _thenElseCondition :: String
-, _thenElseThen :: CaptureUnknownFields (Section a)
-, _thenElseElse :: CaptureUnknownFields (Section a)
+, _thenElseThen :: CaptureUnknownFields (Section a a)
+, _thenElseElse :: CaptureUnknownFields (Section a a)
 } deriving (Eq, Show, Generic)
 
 instance (FromJSON a, HasFieldNames a) => FromJSON (CaptureUnknownFields (ThenElse a)) where
@@ -310,7 +328,7 @@ instance FromJSON BuildType where
     "Custom"    -> return Custom
     _           -> fail "build-type must be one of: Simple, Configure, Make, Custom"
 
-type ExecutableConfig = CaptureUnknownFields (Section ExecutableSection)
+type ExecutableConfig = CaptureUnknownFields (Section ExecutableSection ExecutableSection)
 
 data PackageConfig = PackageConfig {
   packageConfigName :: Maybe String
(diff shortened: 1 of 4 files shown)


scott-fleischman:
Fixes #116

scott-fleischman:
I force-pushed an update to my branch that fixes a couple issues with rendering. I have tried it on our real-world use cases, and they work as expected.

scott-fleischman:
We might consider offering a warning when `exposed-modules` and/or `other-modules` is used within a conditional and one of them has the default value at the root level (meaning it will fill its value recursively).

For example:
```
library:
  source-dirs: src
  exposed-modules: Foo
  when:
    condition: os(windows)
    exposed-modules: Bar
```

will generate:
```
library:
  exposed-modules: Foo
  other-modules:
  - Bar
  - Paths_foo
  if os(windows)
    exposed-modules: Bar
```

Note that it includes `Bar` in `other-modules` outside of the conditional, which is unlikely the desired result.

sol:
@scott-fleischman Hey, on a first glance this looks reasonable.  I will do a proper review when I have time, but it may take a while.

src/Hpack/Config.hs
```diff
@@ -271,18 +271,36 @@ instance FromJSON Condition where
...
   parseJSON = captureUnknownFields
 
-instance HasFieldNames ThenElse
+instance HasFieldNames a => HasFieldNames (ThenElse a)
 
-instance FromJSON ThenElse where
+instance (HasFieldNames a, FromJSON a) => FromJSON (ThenElse a) where
   parseJSON = genericParseJSON_
 
 data Empty = Empty
```
sol: Can `Empty` be removed now?

scott-fleischman: `Empty` is not needed in the library code. It is still used in tests, mainly for the `ToJSON` and `HasFieldNames` instances. I'll move `Empty` to the test `Helper` module.

sol: Cool, makes sense.

Late here already, I'll try to continue the review tomorrow (no promise).

sol:
@scott-fleischman If I specify e.g. `github` in a global conditional then it has no effect, it's just ignored, right?

I'm puzzled whether we want to parametrize `Section` twice, once for it's additional top-level field and once for the fields allowed in a conditional.  We would then use:

- `Section LibrarySection LibrarySection` for libraries
- `Section ExecutableSection ExecutableSection` for executables, tests and benchmarks
- `Section PackageConfig Empty` for the package itself

Any thoughts?

scott-fleischman:
I'll give that approach a try and see how it goes.

scott-fleischman:
@sol I force-pushed an update that uses `Section c d` where `c` is the `Section` data for the conditionals and `d` is the `Section` data for the root section. The `d` so that `Section c` can be a `Functor` over the `sectionData` field.

The code builds, but the tests aren't passing yet. There are a few bugs in the unknown fields code still. But mainly I wanted to get your feedback whether this is the direction you would like this implementation to go. The `Section` code gets messier because of the two type arguments. And I had to define the `Functor` instance for `Conditional` manually.

I kind of prefer the older way due to its simplicity. This way makes the `Section` tree rather subtle in how the conditional type argument gets propagated in the children.

scott-fleischman:
I will also have to rebase my commits.

Just let me know whether you want `Section c d` or whether to revert back to `Section a`.

Thanks!

sol:
@scott-fleischman Sorry for the late response again.  I think we should reject e.g. `github` in top-level conditionals, as that is incorrect.  Unless we find a different way to achieve this, I would go with this approach.  But then again, I haven't written the code and may have the wrong feeling on how messy this approach gets.  Just from a quick glance it looks better than I had imagined.

If you are ok with this approach, please go ahead and rebase.

scott-fleischman:
@sol I finally got the motivation to finish this PR, rebasing off the latest master and fixing the broken tests. Let me know what you think!

scott-fleischman:
I basically just started from the `master` branch and re-applied the same work piece by piece in order to isolate each step (mainly so I could find the source of the previous broken tests).

scott-fleischman:
Personally having two type parameters for `Section` is a bit verbose, but mainly confusing. `Section b a` uses `b` for the conditional branch and `a` for the data field. It's easy to get those backwards, which is what was the source of the earlier test failures.

One consideration is whether we want to add a type synonym for `Section a a` (perhaps `type Section' a = Section a a`), since `Section ExecutableSection ExecutableSection` gets a bit verbose. However, I tend to avoid type synonyms so I didn't include them initially.

scott-fleischman:
I tested the code against our real-world use case and it works great.

scott-fleischman:
In the [cabal project](https://github.com/haskell/cabal), you can see a few uses of these features.

In the Cabal.cabal [here](https://github.com/haskell/cabal/blob/e63c0f3a9db3172536d2f2fff15638fba98364a8/Cabal/Cabal.cabal#L330) and [here](https://github.com/haskell/cabal/blob/e63c0f3a9db3172536d2f2fff15638fba98364a8/Cabal/Cabal.cabal#L457)

In tests: [here](https://github.com/haskell/cabal/blob/e63c0f3a9db3172536d2f2fff15638fba98364a8/cabal-testsuite/PackageTests/Outdated/repo/base-3.0.3.1/base.cabal#L25), [here](https://github.com/haskell/cabal/blob/e63c0f3a9db3172536d2f2fff15638fba98364a8/cabal-testsuite/PackageTests/Outdated/repo/base-3.0.3.2/base.cabal#L31) and [here](https://github.com/haskell/cabal/blob/e63c0f3a9db3172536d2f2fff15638fba98364a8/cabal-testsuite/PackageTests/Outdated/repo/base-4.0.0.0/base.cabal#L24)

scott-fleischman:
@sol as you questioned the validity of fields like `github` in a top-level conditional, I did a cursory investigation of the cabal code and couldn't find a situation where that would be possible in a cabal file. But I did see usages of `other-modules` and `exposed-modules` as I pointed out in my previous comment.

The current code in this PR reflects both of these (omits top-level package-specific fields in conditionals, but supports library and executable-specific fields in conditionals.

scott-fleischman:
@sol At this point, what are your thoughts on this PR and/or the issue it's related to?

sol:
@scott-fleischman I'm very much sorry for being unresponsive on this.  The reason is lack of cycles on my side and that this PR is somewhat biggish.

I'll be busy tomorrow and for the weekend, but I'll try to get back at this early next week (no promise).

src/Hpack/Config.hs
```diff
@@ -275,18 +275,36 @@ instance FromJSON Condition where
 
 instance HasFieldNames Condition
 
-data ThenElse = ThenElse {
+data FlatThen a = FlatThen {
+  _flatThenCondition :: String
+, _flatThenBody :: CaptureUnknownFields (Section a a)
+} deriving (Eq, Show, Generic)
```
sol: What we do here with `FlatThen` is essentially saying: On this level we want to accept all the fields of `Condition` and all the fields of `Section`.  I added a `Product` data type that allows us to solve this kind of situation generically.  So instead of `FlatThen` we can have `Product (Section a a) Condition` now.

src/Hpack/Config.hs
```diff
@@ -440,18 +466,27 @@ data Section a = Section {
 , sectionInstallIncludes :: [FilePath]
 , sectionLdOptions :: [LdOption]
 , sectionBuildable :: Maybe Bool
-, sectionConditionals :: [Conditional]
+, sectionConditionals :: [Conditional b]
 , sectionBuildTools :: Dependencies
-} deriving (Eq, Show, Functor, Foldable, Traversable)
+} deriving (Show)
```
sol: I noticed that in the parsed result we actually don't need two type parameters (we only have `Section Library Library` and `Section Executable Executable`).  The second type parameter is only needed during parsing.

As a second observation: We have types that represent the syntactical structure of `package.yaml` that we use for parsing the actual YAML.  We then map those syntactic types to types that represent a `Package` semantically (e.g. `ExecutableConfig` is mapped to `Executable`).

For `Section` the syntactic type is `CommonOptions`, but here we break the strict separation between syntactic and semantic types.  We use `Section` during parsing the actual YAML as well.

What I did is: Remove the `FromJSON` instance for `Section` and only use `CommonOptions` during parsing.

src/Hpack/Config.hs
```diff
@@ -416,12 +442,12 @@ data Library = Library {
 } deriving (Eq, Show)
 
 data Executable = Executable {
-  executableName :: String
-, executableMain :: FilePath
+  executableName :: Maybe String
```
sol: `executableName` can never be `Nothing` and is never valid inside a conditional.  So what we want here is a type that represents this properly.  There are probably several ways to achieve this, what I did is:

Change `[Section Executable]` to `Map String (Section Executable)` in `Package` and remove `executableName` from `Executable`.  So what we have now is a mapping from executable names to executable sections.

src/Hpack/Config.hs
```diff
@@ -375,7 +393,15 @@ readPackageConfig file = do
...
-      Right <$> mkPackage dir config
+      result@(_, pkg) <- mkPackage dir config
+      case traverse validateExecutable $ packageExecutables pkg ++ packageTests pkg ++ packageBenchmarks pkg of
+        Left err' -> return (Left err')
+        Right _ -> return (Right result)
+      where
+        validateExecutable :: Section Executable Executable -> Either String ()
+        validateExecutable ex = case executableMain . sectionData $ ex of
+          Nothing -> Left "Missing main"
+          Just _ -> Right ()
```
sol: From what I tested, `Cabal` does not require `main` top-level, e.g. the following is a valid executable section:

```
executable foo
  if os(windows)
    buildable: False
  else
    hs-source-dirs: src, driver
    main-is: Main.hs
```

So just omitting `validateExecutable` is fine I guess.

src/Hpack/Config.hs
```diff
@@ -691,37 +729,55 @@ toLibrary dir name globalOptions library = traverse fromLibrarySection sect >>=
...
+          return (Executable (Just name) mainSrcFile modules, ghcOptions)
           where
             inferModules :: IO [String]
             inferModules = filterMain . (++ [pathsModule]) . concat <$> mapM (getModules dir) sectionSourceDirs
 
             pathsModule = pathsModuleFromPackageName packageName_
 
             filterMain :: [String] -> [String]
-            filterMain = maybe id (filter . (/=)) (toModule $ splitDirectories executableSectionMain)
+            filterMain = maybe (const []) (\m -> maybe id (filter . (/=)) (toModule $ splitDirectories m)) executableSectionMain
```
sol: I think `const []` needs to be `id` instead (if there is no `main` we don't filter anything).

src/Hpack/Config.hs
```diff
@@ -691,37 +729,55 @@ toLibrary dir name globalOptions library = traverse fromLibrarySection sect >>=
...
+
+            (mainSrcFile, ghcOptions) = case executableSectionMain of
+              Nothing -> (Nothing, [])
+              Just m -> let (f, o) = parseMain m in (Just f, o)
 
-            (mainSrcFile, ghcOptions) = parseMain executableSectionMain
+        fromExecutableSectionNamed :: ExecutableSection -> Executable
+        fromExecutableSectionNamed (ExecutableSection main other) = Executable Nothing main (fromMaybeList other)
```
sol: I actually infer `other-modules` inside conditional too now (provided the user specified any `source-dirs`).  Any reason we wouldn't want that?

scott-fleischman:
@sol thanks for looking at the PR. It's disappointing there's no record of my changes in the project, but I am glad the issue has a resolution, and with cleaner code than I had.

sol:
> It's  disappointing there's no record of my changes in the project, but I am  glad the issue has a resolution, and with cleaner code than I had.

I'm very much sorry for that.  I tried to retain your authorship but with the amount of refactoring on `master` and the number of rebases this proved more and more laborsome so I eventually gave up on it.  Not quite the same, but I gave you attribution in the CHANGELOG (d3896c1bd553bba1c34e6bdebe93b11402af6) and I'll make sure this won't happen with future contributions.

sol:
> I think that inferring other-modules by default in conditionals will make the default behavior not work very well, which unfortunately goes against one of the values of hpack--if you use the defaults things "just work".
>
> The main issue is that the in a conditional, usually you want to exclude one or more other-modules that will break the build otherwise. That's why I think the preferred way to use conditionals is to manually specify other-modules to avoid getting inferred modules that break that build.
>
> I will see if I can put together some examples that clarify it.


I couldn't come up with any example myself, where not inferring `other-modules` for executables would be more convenient.  So yes, if you can come up with something, that would be very much appreciated!

For libraries, inferring `exposed-modules` / `other-modules` inside conditionals is less straight forward, this is why I deferred that for now.

PR #184, opened 2017-06-22
Recognize alex and happy files as Haskell modules

charles-cooper:
Fixes https://github.com/sol/hpack/issues/183


sol:
Thanks a lot!

Issue #169, opened 2017-04-27
pkgconfig-depends support


KaneTW:
Some C libraries require pkgconfig to properly specify include dirs, e.g. libweston.

sol:
PR please!

sol:
On `master`.

PR #185, opened 2017-06-25
Add support for `pkg-configs` section (#169)

BlakeRain:
Add support for `pkg-configs` in a `Section` which emits a `pkgconfig-depends` (ref issue #169).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: false
 
-language: c
+language: generic
 
 matrix:
   include:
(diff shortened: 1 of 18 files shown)


sol:
@BlakeRain Thanks a lot for working on this!

Regarding naming, I agree that we don't want to use Cabal's `pkgconfig-depends`.  Still, I'm not yet sure whether I like `pkg-confis`.  Other options:

- `pkg-config-dependencies` - conveys meaning and is in line with `dependencies`
- `native-dependencies` - I love this one

Opinions?

We also need to update the README.

BlakeRain:
The `pkg-config-dependencies` does indeed have merit. I'm worried that `native-dependencies` might be confused with `extra-libraries`.

sol:
@BlakeRain `pkg-config-dependencies` or `native-dependencies`, you can make a call.

ip1981:
I think it should include `pkg-config` in name (it relies on [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/), right?), and should not have `native` (it is too vague; for me it's about cross-compilation).

BlakeRain:
This PR seems to have died (probably my fault). Is it worth resurrecting?

sol:
@BlakeRain Yes, please!

krakrjak:
I've sent a PR to @BlakeRain with a merge fixup for his initial patch set. 
https://github.com/krakrjak/hpack/tree/merge-hpack-pr185-iss169

PR #189, opened 2017-08-06
fix table typo

mizunashi-mana:


mizunashi-mana:
Sorry, this is my mistake...

Issue #181, opened 2017-06-14
Ignoring unknown field "executable" in package description


koxu1996:
I have simple **package.yaml** file:
```
name: test
version: 0.1.0.0

dependencies:
  - base >=4.8 && <4.9

library:
  source-dirs: src-lib
  exposed-modules:
    - Game

executable:
  main: Main.hs
  source-dirs: src-exec
  dependencies:
    - test
```
Running `stack build` result in:
`WARNING: Ignoring unknown field "executable" in package description`

I am not sure this is hpack or stack issue. Maybe I am doing something wrong?

sol:
@koxu1996 Thank you for taking the time to open this issue!  The `executable`-field is only support since `hspack-0.18.0`, which is not used by `stack` yet.   For now you may want to use

```yaml
executables:
  test:
    main: Main.hs
    source-dirs: src-exec
    dependencies:
      - test
```
instead.

PR #191, opened 2017-08-19
Mention version requirement

chrissound:
Could we mention the version requirement? 
https://github.com/sol/hpack/issues/181


sol:
This information will inevitably get outdated in the future.  And I'm not willing to take the responsibility to keep it up-to-date.

What we could consider is to add some general advice for users to always use the latest version of `hpack`.  Please feel free to open a separate PR if you think this is a good idea.

Issue #91, opened 2016-04-22
Support npm-style version ranges?


tfausak:
I would love to be able to specify version numbers using npm's [advanced range syntax](https://docs.npmjs.com/misc/semver#advanced-range-syntax). Stuff like `1.2.x`, `~1.2.3`, and `^1.2.3`. I think the [semver-range](http://hackage.haskell.org/package/semver-range-0.1.1/docs/Data-SemVer.html) package can represent them, but I haven't ever played with it.

soenkehahn:
Btw, `cabal` (and `hpack`) already support e.g. `1.2.*`. Which I guess is equal to `1.2.x`.

tfausak:
That's true. I'm more interested in the tilde and caret specifiers.

mitchellwrosen:
I'm willing to implement this but I want to make sure it's a welcome change. What do you think, @sol?

sol:
Skimming through the documentation, I don't think you can translate this to Haskell directly. It seems that Node always uses major, minor, patch. This is not true for the Haskell. Firstly the pvp defines the major version as `x.y`, secondly you see a mixture of versions like `2`, `2.1`, `2.1.1`, `2.0.0.1` on Hackage.

So if somebody thinks it's worth working on this, the first step would probably be to write down how exactly we would translate this to Haskell.

It may also make sense to consider each feature individually. E.g. I think version ranges could be nice, if we can figure out the semantics. On the other hand I'm not excited about `.x` as we already have star syntax.

mitchellwrosen:
Hackage does allow e.g. `2`, but to the cabal solver, this is the same as `2.0.0.0`, no?

I think the overall ideas do translate directly to Haskell, we'd just have to account for major versions taking up two slots rather than one.

Agreed that `.x` doesn't make much sense, and it would be best to start with just one, like `1.0 - 2.0`

soenkehahn:
IIRC to `cabal` `2` is not the same as `2.0.0.0`. I think the constraint `<= 2` would even disallow `2.0.0.0` because `Data.Version` says that `2 < 2.0.0.0`. I may be wrong, though.

sol:
I have some similar resemblance of something fishy there

sol:
As much as I like this idea, I think it's not straight forward to apply this to Cabal (see the comments of @soenkehahn an me above).

If somebody still wants aspects of this to happen (e.g. version ranges), then please feel free to open individual tickets for that.

PR #193, opened 2017-08-30
Allow specifying dependencies as a hash

### LINKED ISSUE #186: Allow dependencies to be a hash?
Currently the `dependencies` field must be either a single element or a list of elements. Each element is either a string (like `'base == 4.9.1.0'`) or a hash (like `{ name: foo, path: ./foo }`). I think it would be nice to specify dependencies as a hash where the key is the package name and the value is the version constraint. For example:

``` yml
dependencies:

  # implies ==
  base: 4.9.1.0

  # wildcards
  bytestring: 0.10.*

  # comparisons
  text: '>= 1.2.2'

  # operators
  vector: '>= 0.11 && < 0.13'

  # no constraints
  deepseq: null

  # local, implies { name: foo }
  foo: { path: ./foo }

  # git, also implies name
  bar: { git: 'http://example.com', ref: master }
```

This would create ambiguity between a single dependency given as a hash and my proposed syntax. I would resolve that ambiguity by forbidding a single dependency to be given as a hash. In other words, to list a single local or Git dependency, you'd have to do this:

``` yml
dependencies: [ { name: foo, path: ./foo } ]
```

Without the list wrapping the hash, it would be parsed like this:

``` yml
dependencies: { name: foo, path: ./foo }
# dependencies: [ 'name == foo', 'path == ./foo' ]
```

If you think this is worth exploring I can probably come up with a PR to add it.

tfausak:
This pull request fixes #186. It changes the dependencies to allow hash values specifying the package as the key and the constraint as the value. Here are some allowable dependencies:

``` yaml
# As before, a single string is allowed.
dependencies: base ==4.10.0.0

# Also as before, a list of strings is allowed.
dependencies:
  - base ==4.10.0.0

# This changed! Now a hash is interpreted as "<package>: <constraint>".
dependencies:
  base: ==4.10.0.0

# Values aren't required.
dependencies:
  base:
  aeson: null

# Values can be hashes.
dependencies:
  local:
    path: the/path-to/local
  remote:
    github: user/repo
    ref: master

# Be careful with constraints that collide with YAML syntax.
dependencies:
  # base: >=4.10 # Error!
  base: '>=4.10'

# When there are duplicate keys, later keys win.
dependencies:
  base: ignored
  base: ==4.10.0.0
```

In #186 I originally suggested that the `==` constraint be implied if no constraints were given. In other words, `base: 4.10.0.0` would mean `base ==4.10.0.0`. I chose not to implement that because it would require parsing constraints, which is hard.

### DIFF
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -1,4 +1,4 @@
--- This file has been generated from package.yaml by hpack version 0.18.0.
+-- This file has been generated from package.yaml by hpack version 0.18.1.
 --
 -- see: https://github.com/sol/hpack
 
@@ -39,6 +39,7 @@ library
     , unordered-containers
     , yaml
     , aeson >= 0.11
+    , vector
   exposed-modules:
       Hpack
       Hpack.Config
@@ -72,6 +73,7 @@ executable hpack
     , unordered-containers
     , yaml
     , aeson >= 0.11
+    , vector
     , hpack
   default-language: Haskell2010
 
@@ -96,6 +98,7 @@ test-suite spec
     , unordered-containers
     , yaml
     , aeson >= 0.11
+    , vector
     , hspec == 2.*
     , QuickCheck
     , temporary
--- a/package.yaml
+++ b/package.yaml
@@ -24,6 +24,7 @@ dependencies:
   - unordered-containers
   - yaml
   - aeson >= 0.11
+  - vector
 
 library:
   source-dirs: src
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -404,31 +404,37 @@ instance FromJSON Dependency where
     Object o -> addSourceDependency o
     _ -> typeMismatch "String or an Object" v
     where
-      addSourceDependency o = Dependency <$> name <*> (Just <$> (local <|> git))
+      addSourceDependency o = Dependency <$> name <*> (Just <$> source)
         where
           name :: Parser String
           name = o .: "name"
 
-          local :: Parser AddSource
-          local = Local <$> o .: "path"
-
-          git :: Parser AddSource
-          git = GitRef <$> url <*> ref <*> subdir
-
-          url :: Parser String
-          url =
-                ((githubBaseUrl ++) <$> o .: "github")
-            <|> (o .: "git")
-            <|> fail "neither key \"git\" nor key \"github\" present"
-
-          ref :: Parser String
-          ref = o .: "ref"
-
-          subdir :: Parser (Maybe FilePath)
-          subdir = o .:? "subdir"
+          source :: Parser AddSource
+          source = parseJSON v
 
 data AddSource = GitRef GitUrl GitRef (Maybe FilePath) | Local FilePath
   deriving (Eq, Show, Ord)
+
+instance FromJSON AddSource where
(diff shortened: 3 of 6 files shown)


sol:
@tfausak Hey, sorry for the late reply (I was offline for a while).

> In #186 I originally suggested that the == constraint be implied if no constraints were given. In other words, base: 4.10.0.0 would mean base ==4.10.0.0. I chose not to implement that because it would require parsing constraints, which is hard.

This syntax is actually what convinced me in favor of adding this feature.  In addition, I was hopping that it will be possible to tighten (or loosen) constraints for specific dependencies, e.g.:

```yaml
dependencies:
  base: 4.*

tests:
  ...
  dependencies:
    base: '>= 4.3'
```

What exactly do you mean with "hard" when it comes to parsing constraints?  My assumption was that depending on `Cabal` will allow us to do this (not that I'm excited to depend on `Cabal`, but that's probably what I would do...).

tfausak:
I said that parsing constraints would be hard because right now the dependency strings are simply passed through to the Cabal file. Parsing the constraints will be a little bit bigger of a change, but it's definitely doable. 

Would later constraints be "and"ed together with earlier ones? In your example, that would end up as `4.* && >=4.3`. It would be hard to loosen constraints that way. You could optionally put `||` before the constraint, but Cabal might not like parsing that.

sol:
> Would later constraints be "and"ed together with earlier ones? In your example, that would end up as 4.* && >=4.3.

Cabal solves dependencies for a package as a whole (with intersection-semantics for constraints from different sections, which is the same as `&&`).  For that reason I think it is fine to override global constraints with section specific ones by just merging the `Map`s and leaving the intersecting to the dependency solver.

This will give us the exact same semantics in most cases.  The only exception I see is when you **override a global dependency in all sections of a package**, e.g.:

```yaml
dependencies:
  base: '>= 4'

library:
  dependencies:
    base: '< 5'
  ...

tests:
  spec:
    dependencies:
      base: '>= 4.3'
```

which is not particularly DRY in the first place and better expressed as

```yaml
dependencies:
  base: 4.*

tests:
  spec:
    dependencies:
      base: '>= 4.3'
```

(As a side note, as I understand it, it's never useful to specify different constraints for `library` and `executables`, as Cabal will always solve them together anyway.  In the same vane it does not make sense to loosen constraints for `tests`/`benchmarks`, as again Cabal will always solve them together with constraints from `library`/`executables`.  So the only thing that I ever wanted to do is tighten constraints for `tests`/`benchmarks`.  Please correct me if you think I'm missing something.)

> It would be hard to loosen constraints that way. You could optionally put || before the constraint, but Cabal might not like parsing that.

Can you come up with a use case?

@tfausak big sorry again for the delay!  I implemented constraint parsing and changed the internal representation from a list to a `Map` in #196.  Can you migrate your changes on top of that?

sol:
Once we are through with this, I'm also happy to revisit #91 and see what we can cherry-pick from npm/bundler.

tfausak:
I'll take another look at this today. Your changes look great and lay a nice foundation for what I want to do here. I'll probably open another PR to avoid gnarly merge conflicts.

tfausak:
Closing in favor of #198.

PR #197, opened 2017-09-24
Upgrade to Stackage LTS 9.0

tfausak:
This upgrades the Stack configuration from GHC 7.10.2 to GHC 8.0.2. It also upgrades a whole slew of packages. Everything built fine on my machine. 

I ran into this while working on a fix for #186 because I needed a more recent version of Cabal to parse constraints in isolation.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,1 @@
-flags: {}
-packages:
-- '.'
-extra-deps:
-- aeson-0.11.0.0
-- yaml-0.8.21.2
-resolver: lts-5.3
+resolver: lts-9.0


sol:
@tfausak Cool, I'll merge as soon as the build is green.

As a side note, you will be affected by https://ghc.haskell.org/trac/ghc/ticket/13285 if you use GHC 8.0.*.

sol:
(as we use sections a lot in the test suite)

tfausak:
Thanks! I'll look out for that. It hasn't bitten me yet. 

Turns out I didn't actually need this upgrade for my work. I was confused about where `DependencyVersion` came from 😄 Regardless, this would be nice to have.

sol:
Actually, can we go with `lts-6.35` or the latest nightly instead?

tfausak:
Sure, either one is fine with me. I went with `lts-9.0` because that's the one I use for most of my projects so it had the highest cache hit rate.

PR #198, opened 2017-09-24
Allow specifying dependencies as a hash

### LINKED ISSUE #186: Allow dependencies to be a hash?

tfausak:
This fixes #186. It's my second attempt after #193. 

This pull request changes the dependencies to allow hash values specifying the package as the key and the constraint as the value. Here are some allowable dependencies:

``` yaml
# As before, a single string is allowed.
dependencies: base ==4.10.0.0

# Also as before, a list of strings is allowed.
dependencies:
  - base ==4.10.0.0

# This changed! Now a hash is interpreted as "<package>: <constraint>".
dependencies:
  base: ==4.10.0.0

# Values aren't required.
dependencies:
  base:
  aeson: null

# Values can be hashes.
dependencies:
  local:
    path: the/path-to/local
  remote:
    github: user/repo
    ref: master

# Be careful with constraints that collide with YAML syntax.
dependencies:
  # base: >=4.10 # Error!
  base: '>=4.10'

# When there are duplicate keys, later keys win.
dependencies:
  base: ignored
  base: ==4.10.0.0
```

There are two things that I have not done that I would like to:

- Parse `base: 4.10.0.0` as `base ==4.10.0.0`. That is, use `==` as the constraint if none is given. I haven't done this yet because it will require parsing `DependencyVersion` values into actual `VersionRange`s.

- Merge dependencies from the `library` section into other sections. If you have `library: { dependencies: base ==4.8.* }` and `tests: { test: { dependencies: base >=4.8.2 } }`, the test's dependency on base should be `==4.8.* && >=4.8.2`. I haven't done this yet because it will require mucking around in `mkPackage`.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -21,7 +21,7 @@ module Hpack.Config (
 , package
 , section
 , Package(..)
-, Dependencies
+, Dependencies(..)
 , Dependency(..)
 , DependencyVersion(..)
 , SourceDependency(..)
@@ -108,7 +108,7 @@ renamePackage name p@Package{..} = p {
   }
 
 renameDependencies :: String -> String -> Section a -> Section a
-renameDependencies old new sect@Section{..} = sect {sectionDependencies = (Map.fromList . map rename . Map.toList) sectionDependencies, sectionConditionals = map renameConditional sectionConditionals}
+renameDependencies old new sect@Section{..} = sect {sectionDependencies = (Dependencies . Map.fromList . map rename . Map.toList . unwrapDependencies) sectionDependencies, sectionConditionals = map renameConditional sectionConditionals}
   where
     rename dep@(name, version)
       | name == old = (new, version)
@@ -124,7 +124,7 @@ packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . d
   ++ (concatMap deps packageBenchmarks)
   ++ maybe [] deps packageLibrary
   where
-    deps xs = [Dependency name version | (name, version) <- (Map.toList . sectionDependencies) xs]
+    deps xs = [Dependency name version | (name, version) <- (Map.toList . unwrapDependencies . sectionDependencies) xs]
 
 section :: a -> Section a
 section a = Section a [] mempty [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty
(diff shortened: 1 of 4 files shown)


sol:
@tfausak This looks great.  I'll add some comments, but all of them are more or less bike-shedding.

src/Hpack/Config.hs
```diff
@@ -231,12 +233,12 @@ instance HasFieldNames ExecutableSection
 instance FromJSON ExecutableSection where
   parseJSON = genericParseJSON_
 
-dependenciesFromList :: Maybe (List Dependency) -> Map String DependencyVersion
-dependenciesFromList xs = Map.fromList [(name, version) | Dependency name version <- fromMaybeList xs]
+dependenciesFromList :: Maybe Dependencies -> Dependencies
+dependenciesFromList = fromMaybe mempty
```
sol: Arguably, this is not a good name for this function anymore.  I would tend to just inline it.

src/Hpack/Config.hs
```diff
@@ -397,30 +399,39 @@ instance FromJSON Dependency where
...
-          ref = o .: "ref"
+newtype Dependencies = Dependencies {
+  unwrapDependencies :: Map String DependencyVersion
+} deriving (Eq, Show)
 
-          subdir :: Parser (Maybe FilePath)
-          subdir = o .:? "subdir"
+instance Monoid Dependencies where
+  mempty = Dependencies mempty
+  mappend (Dependencies x) (Dependencies y) = Dependencies $ mappend x y
```
sol: The `Monoid` instance can be derived.

src/Hpack/Config.hs
```diff
@@ -397,30 +399,39 @@ instance FromJSON Dependency where
...
-          url :: Parser String
-          url =
-                ((githubBaseUrl ++) <$> o .: "github")
-            <|> (o .: "git")
-            <|> fail "neither key \"git\" nor key \"github\" present"
-
-          ref :: Parser String
-          ref = o .: "ref"
+newtype Dependencies = Dependencies {
+  unwrapDependencies :: Map String DependencyVersion
```
sol: `unDependencies` would be a bit more idiomatic.

sol:
We need a changelog entry and ideally also documentation.

tfausak:
:tada: Thanks for merging this patch and making those changes!

PR #199, opened 2017-09-26
Accept specific versions and wildcard versions for dependencies

sol:


sol:
@tfausak any opinion?

PR #200, opened 2017-09-27
Internal libraries

bucklereed:
I have very tentatively put this as 'since 0.20', since I see that there's an 0.19 on Hackage, but the changelog still lists that as 'next' and there's no 0.19 tag in the repository...

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ existing cabal file into a `package.yaml`.
 | `custom-setup` | · | | See [Custom setup](#custom-setup) | | |
 | `flags`  | `flag <name>` | | Map from flag name to flag (see [Flags](#flags)) | | |
 | `library` | · | | See [Library fields](#library-fields) | | |
+| `internal-libraries` | `library <name>` | | Map from internal library name to a dict of [internal library fields](#internal-library-fields) and global top-level fields. | | `0.20.0` |
 | `executables` | `executable <name>` | | Map from executable name to executable (see [Executable fields](#executable-fields)) | | |
 | `executable` | `executable <package-name>` | | Shortcut for `executables: { package-name: ... }` | | `0.18.0` |
 | `tests` | `test-suite <name>` | | Map from test name to test (see [Test fields](#test-fields)) | | |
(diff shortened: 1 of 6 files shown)


sol:
@bucklereed Hi, thanks for taking the time open this PR and sorry for the delay.

The approach you took looks good to me.  However, I want to resolve #175 first; let's revisit this after.

sol:
@bucklereed #175 is done, so let's tackle this.  Can you rebase your code on `master`?

bucklereed:
Rebased and should be ready to go, Travis willing.

src/Hpack/Config.hs
```diff
@@ -212,6 +214,20 @@ instance HasFieldNames LibrarySection
 instance FromJSON LibrarySection where
   parseJSON = genericParseJSON_
 
+data InternalLibrarySection = InternalLibrarySection {
+  internalLibrarySectionExposedModules :: Maybe (List String)
+, internalLibrarySectionOtherModules :: Maybe (List String)
+, internalLibrarySectionReexportedModules :: Maybe (List String)
+} deriving (Eq, Show, Generic)
```
sol: The only difference between `InternalLibrarySection` and `LibrarySection` is that `InternalLibrarySection` won't accept `exposed`.  However, from what I tried, `Cabal` will accept `exposed` for internal libraries.  I'm not sure what exactly the effect is (if any).  But for code simplicity, let's follow what cabal does and use `LibrarySection` instead of `InternalLibrarySection`.

src/Hpack/Config.hs
```diff
@@ -433,6 +451,12 @@ data Library = Library {
 , libraryReexportedModules :: [String]
 } deriving (Eq, Show)
 
+data InternalLibrary = InternalLibrary {
+  internalLibraryExposedModules :: [String]
+, internalLibraryOtherModules :: [String]
+, internalLibraryReexportedModules :: [String]
+} deriving (Eq, Show)
```
sol: Let's use `Library` instead of `InternalLibrary`.

src/Hpack/Config.hs
```diff
@@ -759,6 +791,28 @@ mapSectionAcc f = go
       , sectionConditionals = map (go acc <$>) sectionConditionals
       }
 
+toInternalLibraries :: FilePath -> String -> Section global -> [(String, Section InternalLibrarySection)] -> IO ([String], Map String (Section InternalLibrary))
```
sol: I guess we can get rid of this too and use `toLibrary` instead.

README.md
```diff
@@ -102,6 +103,15 @@ These fields are merged with all library, executable, test, and benchmark compon
...
 
+#### <a name="internal-library-fields"></a>Internal library fields
+
+| Hpack | Cabal | Default | Notes |
+| --- | --- | --- | --- |
+| `exposed-modules` | · | All modules in `source-dirs` less `other-modules` | |
+| `other-modules` | · | All modules in `source-dirs` less `exposed-modules` | |
+| `reexported-modules` | · | | |
+| | `default-language` | `Haskell2010` | |
+
```
sol: Given the other changes, we can get rid of this too and reference `library-fields` instead.

bucklereed:
OK. I've prepared a new branch with those changes and I'll open a new PR for that in a second, because apparently Github's pull requests are fixed to one branch forever.

PR #211, opened 2017-11-05
Add support for user-defined extension sets

sol:
Example usage:

```yaml
name: hello
defaults: sol-2017

library:
  ..
```

TODO:
- [x] Mock HTTP requests in tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,50 @@ existing cabal file into a `package.yaml`.
 | `executable` | `executable <package-name>` | | Shortcut for `executables: { package-name: ... }` | | `0.18.0` |
 | `tests` | `test-suite <name>` | | Map from test name to test (see [Test fields](#test-fields)) | | |
 | `benchmarks` | `benchmark <name>` | | Map from benchmark name to benchmark (see [Benchmark fields](#benchmark-fields)) | | |
+| `defaults` | | | See [Defaults](#defaults) | | |
 
-Hpack does not require you to specify a `cabal-version` manually, it is
-automatically set depending on which features are used.
+**Note:** Hpack does not require you to specify a `cabal-version` manually.  When
+generating a `.cabal` file, Hpack sets the `cabal-version` automatically based
+on the features that are used.
+
+#### <a name="defaults"></a>Defaults
+
+Hpack allows the inclusion of [common fields](#common-fields) from a file on
+GitHub.
+
+To use this feature a user must specify a GitHub repository, Git reference and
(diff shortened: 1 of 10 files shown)


bucklereed:
I have some concerns about this.

1. It's tied to Github. Please consider alternative hosting providers, as well as having locally defined extension sets that aren't up on the web.

2. Build tools shouldn't need network access to do their jobs. This has benefits for reproducibility, archiveability, privacy and reliability. It's okay for hpack's default configuration to hit the network, but it should be possible to pass a `--no-network` flag and have it still be usable, which entails having (and documenting) a way to tell it something like 'if you want to access https://blah.example, it can be found in the file /foo/bar'.

sol:
@bucklereed thanks a lot for voicing your concerns and sorry for the late reply.

> It's tied to Github. Please consider alternative hosting providers

I see your point.  While I don't intend to add support for a wide variety of sources upfront, I'm definitely open to add more options as demand for them growths.

My general approach with `hpack` is to "defer features to the point at which we actually want to use them and never add anything that we don't need right now".  This is meant to prevent code bloat and to keep us focused on the things that matter most.  As `hpack` gets more popular it's getting harder to follow this approach strictly, but my basic approach is still: For a feature to be considered there needs to be at least one user who wants to use it.

> as well as having locally defined extension sets that aren't up on the web.

Yes, I agree.  Not part of this PR, but we can extend it to accept a single `path` to include locally defined extension sets, e.g.:

```yaml
defaults:
  path: ../common.yaml
```

> Build tools shouldn't need network access to do their jobs. This has benefits for reproducibility, archiveability, privacy and reliability.

`hpack` will only access the network the first time it sees a specific extension set.  Extension sets are cached under `~/.hpack/defaults/` for later use.  Does this address your concern regarding reproducibility?

As a side note, this also means that if you use `master` as `ref` then `hpack` will never update the extension set to the latest version.  For this reason (and also to guarantee reproducibility across machines) it's recommended to only use tags as `ref`.

> It's okay for hpack's default configuration to hit the network, but it should be possible to pass a --no-network flag and have it still be usable, which entails having (and documenting) a way to tell it something like 'if you want to access https://blah.example, it can be found in the file /foo/bar'.

This is already possible by manually adding the corresponding file for an extension set to `~/.hpack/defaults/`.  If we document that, would that address you concerns?

sol:
@bucklereed I added documentation on how to prevent Hpack from downloading defaults.  If there is still something that you would want to see in Hpack please open a separate issue ;).

Issue #213, opened 2017-11-08
cabal-version is not set properly when using ^>= bounds


quasicomputational:
When `^>=` is used, cabal-version needs to be at least 2.0. Hpack doesn't check for this, though.

sol:
A PR would be more than welcome.

tfausak:
For what it's worth, Cabal itself doesn't check for this either. If you run `cabal check` against a package description with `^>=` bounds and `cabal-version: >= 1.10`, it will succeed.

sol:
What we can do is parse that syntax and `^>=` and expand it as per the semantics given at https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-build-depends.

Independent from that, I think we want some syntactic sugar for version ranges, where he lower bound is inclusive and the upper bound is exclusive.  Stuff like `base >= 4.3.0.0 && < 5` is so common that I think it deserves dedicated support.  I'm  just not sure what the syntax would be, maybe

```yaml
foo: x ~ y  # this desugars to ">= x && < 5"
```

If anybody has input on this, please comment!

tfausak:
Adding custom operators reminds me of #91. I don't think it's a bad idea per se, but it could get tricky if Cabal ends up using the same operator for something else. I think it would be better to lobby for the change in Cabal, then add it to hpack for users that aren't on the latest Cabal.

PR #216, opened 2017-11-19
Support custom cabal-version field

mizunashi-mana:
related #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ existing cabal file into a `package.yaml`.
 | `license-file` | `license-file` or `license-files` | `LICENSE` if file exists | May be a list | | |
 | `tested-with` | · | | | | |
 | `build-type` | · | `Simple`, or `Custom` if `custom-setup` section exists | Must be `Simple`, `Configure`, `Make`, or `Custom` | | |
-| | `cabal-version` | `>= 1.10` or `>= 1.21` | `>= 1.21` if library component has `reexported-modules` field | | |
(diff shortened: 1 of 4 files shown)


sol:
@mizunashi-mana thank you for taking the time to open this PR.

With the current state of the discussion I don't see the need for this change.  Generally, I know of two cases when you wan't to specify a Cabal version:

1. You have a custom `Setup.hs` that requires a certain version of the `Cabal` library.  In that case you can add something like
   ```yaml
   custom-setup:
     dependencies: Cabal == 2.*
   ```
   to `package.yaml`.
1. Your generated cabal file uses features that require a specific version of `Cabal` / `cabal-install`.  In this case it is the responsibility of `hpack` to add the appropriate `cabal-version` to the cabal file when generating it.  If that does not happen, e.g. as with #213, then thats at bug and we rather fix that bug in `hpack` than requiring users of `hpack` to figure out what `cabal-version` is needed and setting it manually.

Unless somebody can give a use case that does not fall in one of this categories, this is a `wontfix`.

mizunashi-mana:
@sol I see.  Should I close this PR?

quasicomputational:
FWIW I think it might be worth having an escape hatch here so that, if hpack is getting it wrong, you're not stuffed until you can get a PR merged here, then wait for stack to make a release that includes a hpack release with the fix. But, any use of such an escape hatch ought to be regarded as a work-around for a bug in hpack, I agree.

sol:
I propose https://github.com/sol/hpack/issues/262 as a way to workaround issues like this.

sol:
Closing.

PR #218, opened 2017-11-26
Fix some for AppVoyer fails

mizunashi-mana:


test/Hpack/ConfigSpec.hs
```diff
@@ -327,7 +327,7 @@ spec = do
           dependencies: ghc-prim
           baz: 42
         |]
-        (`shouldBe` [
+        (`shouldMatchList` [
```
sol: I pushed this to `master`.

src/Hpack/Util.hs
```diff
@@ -159,3 +160,6 @@ type Hash = String
 
 sha256 :: String -> Hash
 sha256 c = show (hash (encodeUtf8 c) :: Digest SHA256)
+
+linesForAll :: String -> [String]
+linesForAll = lines . filter (/= '\r')
```
sol: I think we need a more principled solution for this.  See also https://github.com/sol/hpack/issues/219.

mizunashi-mana: I see.

mizunashi-mana:
OK, I close this PR.

PR #220, opened 2017-12-02
Use system specific newline encoding when doing IO (close #219)

### LINKED ISSUE #219: Use system specific newline encoding when doing IO
As a side effect of https://github.com/sol/hpack/pull/143 we lost system specific newline encoding.

sol:


sol:
@snoyberg @tfausak any input on this?

src/Hpack/Utf8.hs
```diff
@@ -0,0 +1,61 @@
...
+readFile = fmap decodeText . B.readFile
+
+writeFile :: FilePath -> String -> IO ()
+writeFile name xs = withFile name WriteMode (`hPutStr` xs)
+
+putStr :: String -> IO ()
+putStr = hPutStr stdout
+
+hPutStrLn :: Handle -> String -> IO ()
+hPutStrLn h xs = hPutStr h xs >> hPutStr h "\n"
```
tfausak: This should write `"\r\n"` when appropriate, right?

sol: @tfausak `hPutStr` will take care of converting `"\n"` to "\r\n" if necessary.

tfausak: Oops, right!

sol:
Thanks for the feedback guys!

PR #221, opened 2017-12-03
Omit inference of modules that are referenced in conditionals (fixes #209)

### LINKED ISSUE #209: Omit inference of modules that are referenced in conditionals
I propose that any modules appearing in `exposed-modules` and/or `other-modules` within a conditional are excluded from consideration when inferring `exposed-modules` and/or `other-modules` at the top level.

For me, the intuition behind this proposal is that any module in a conditional is likely problematic outside of the conditional (platform-specific or is associated with dependencies in the conditional).

I believe this would best the best default setup that would "just work" for most cases.

This is a continuation of the discussion from https://github.com/sol/hpack/commit/6a55600c5fd86be5ae95eb2f7b217c898519a676

sol:


sol:
Fixes #209.

sol:
@scott-fleischman this is what I came up with?  Can you give this a quick code review?

scott-fleischman:
This works as expected for my use case. I will take a closer look at the code directly.

PR #223, opened 2017-12-05
make github field accept an object

quasicomputational:
This is preferable to inventing a syntax for subdirectories, branches, etc ourselves, with all the benefits of being able to name the parts.

There's a behaviour change as noted in the changelog.

Maybe `branch` should be renamed `ref`. `git` fields could be made to accept similar objects, but this isn't done (yet).

I have to figure out how to get the generic machinery to do what I want, I think, rather than a hand-written FromJSON instance, if I want warnings?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
   - `extra-doc-files` requires setting `cabal-version` to at least
     1.18; this is now done properly.
   - Accept bool for `condition` (see #230)
+  - Allow `github` to be an object, which newly permits setting the branch
+    and the homepage anchor, as well as the subdirectory.
+
+    This changes the generated homepage of any project using `github`
+    with a subdirectory: previously it pointed to the repository's
+    root, but now it points into the subdirectory.
 
 ## Changes in 0.21.2
   - Fix a bug in module inference for conditionals (see #236)
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ existing cabal file into a `package.yaml`.
 | `description` | · | | | | |
 | `category` | · | | | | |
 | `stability` | · | | | | |
-| `homepage` | · | If `github` given, `<repo>#readme` | | | |
+| `homepage` | · | If `github` given, `<repo>(/tree/<branch>)(/<subdir>)(#<anchor>)` | | | |
 | `bug-reports` | · | If `github` given, `<repo>/issues` | | | |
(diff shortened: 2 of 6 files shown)


quasicomputational:
I have hacked this into shape and got it using the generic machinery properly (I hope).

quasicomputational:
Well, that's a fun test failure: it worked on my machine! I guess that it's locale-dependent or something similar, and that, to fix it properly, the warnings are going to have to be sorted by someone somewhere.

sol:
@quasicomputational sorry for getting back late.  Can you motivate this a bit?  Do you have any projects you want to use this on?  We don't necessarily replicate every arcane Cabal feature, only things that we actually want to use right now.

My main "concern" with `branch` is that I just can't see how an efficient workflow would look like for this.  Should I change it before a release to a specific tag? Then change it back after?  Or just leave the latest release on master?  And how the hell is this supposed to be DRY?  If I use this feature then I effectively specify my version in 3 places:

- by `git tag <version>`
- in the `version` field
- in the `source-repository` section

So far I just dismissed the fact that Cabal allows you to specify a branch as over engineered.  I think Darcs, which doesn't have the concept of lightweight branches, might be the historic reason for this (specifically, you can specify multiple source repositories, e.g. `development` and `stable`, which with Darcs would have different URLs, same for `bzr`).

But again, if you really want to use it we can think about how to leverage it.

quasicomputational:
I would have use of this.

The current parsing of `github` fields doesn't allow for multiple levels of subdirectories, and, even if it did, it's sufficiently unclear that I'd want to leave a comment in the code about it: at a glance, I don't think I'd know what `github: foo/bar/baz/qux` meant, but `github: { repo: foo/bar, path: baz/qux }` is crystal clear.

Second, with specifying the homepage through `github`, `#readme` is always appended, which I think isn't always desirable. (I certainly don't desire it!) With this PR, that's configurable too.

RE `branch`: that can go if it's a problem. The idea is to point it at the 'default branch' for the repository, not a tagged version, but 99.5% of the time that's just going to be `master`. It's mostly only there because the branch appears in the Github URLs if you're using a subdirectory and I didn't want to make an arbitrary un-overrideable decision, but I think that always setting it to `HEAD` will probably work.

There would be some use for a `tag` field if we were generating `source-repository this`, but that'd be another feature altogether, I think.

quasicomputational:
I'm not going to get this over the finish line.

PR #224, opened 2017-12-05
Implement extra-doc-files.

### LINKED ISSUE #206: Support `extra-doc-files`
Cabal since 1.18 supports `extra-doc-files`.

http://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-extra-doc-files

I have a few projects that used them before I converted them to `hpack`. Now they're listed in `extra-source-files` to get them in the `sdist` tarball, but it would be nice if `hpack` supported this field as those files are meant to be installed in a package's `html` docs directory.

quasicomputational:
Closes #206.


quasicomputational:
Thanks--I was going to get around to this, but you got there ahead of me!

PR #227, opened 2017-12-05
Implement generated-other/exposed-modules.

### LINKED ISSUE #207: Feature request: support `autogen-modules` field
[autogen-modules](https://cabal.readthedocs.io/en/latest/developing-packages.html#autogenerated-modules) is present in `Cabal-2.0.0.2`, and is useful for representing modules generated by custom `Setup.hs` scripts.

quasicomputational:
These will populate other/exposed-modules as well as
autogen-modules. Testing has revealed that Paths_* does not have to be
listed as a generated module - which is just as well, since using
autogen-modules forces cabal-version >= 2.0.

Closes #207.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
   - `extra-doc-files` requires setting `cabal-version` to at least
     1.18; this is now done properly.
   - Accept bool for `condition` (see #230)
+  - `generated-exposed-modules` and `generated-other-modules`, for populating
+    the `autogen-modules` field (#207).
 
 ## Changes in 0.21.2
   - Fix a bug in module inference for conditionals (see #236)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   - Accept bool for `condition` (see #230)
   - `generated-exposed-modules` and `generated-other-modules`, for populating
     the `autogen-modules` field (#207).
+  - Corrected `cabal-version` setting for `reexported-modules` inside
+    a conditional.
 
 ## Changes in 0.21.2
   - Fix a bug in module inference for conditionals (see #236)
--- a/README.md
+++ b/README.md
@@ -155,11 +155,13 @@ values are merged with per section values.
 
 #### <a name="library-fields"></a>Library fields
 
-| Hpack | Cabal | Default | Default inside conditional |
+| Hpack | Cabal | Default | Notes |
 | --- | --- | --- | --- |
 | `exposed` | · | | |
 | `exposed-modules` | · | All modules in `source-dirs` less `other-modules` less any modules mentioned in `when` | |
-| `other-modules` | · | All modules in `source-dirs` less `exposed-modules` less any modules mentioned in `when` | Only if `exposed-modules` is not specified inside the conditional: All modules in `source-dirs` of the conditional less any modules mentioned in `when` of the conditional |
+| `generated-exposed-modules` | | | Added to `exposed-modules` and `autogen-modules`. Since `0.22.0`.
+| `other-modules` | · | Outside conditionals: All modules in `source-dirs` less `exposed-modules` less any modules mentioned in `when`. Inside conditionals, and only if `exposed-modules` is not specified inside the conditional: All modules in `source-dirs` of the conditional less any modules mentioned in `when` of the conditional | |
+| `generated-other-modules` | | | Added to `other-modules` and `autogen-modules`. Since `0.22.0`.
 | `reexported-modules` | · | | |
 | `signatures` | · | | |
 | | `default-language` | `Haskell2010` | |
@@ -170,6 +172,7 @@ values are merged with per section values.
 | --- | --- | --- | --- |
 | `main` | `main-is` | | |
 | `other-modules` | · | All modules in `source-dirs` less `main` less any modules mentioned in `when` | |
+| `generated-other-modules` | | | Added to `other-modules` and `autogen-modules`. Since `0.22.0`.
 | | `default-language` | `Haskell2010` | |
 
 #### <a name="test-fields"></a>Test fields
@@ -179,6 +182,7 @@ values are merged with per section values.
 | | `type` | `exitcode-stdio-1.0` | |
 | `main` | `main-is` | | |
 | `other-modules` | · | All modules in `source-dirs` less `main` less any modules mentioned in `when` | |
+| `generated-other-modules` | | | Added to `other-modules` and `autogen-modules`. Since `0.22.0`.
 | | `default-language` | `Haskell2010` | |
 
 #### <a name="benchmark-fields"></a>Benchmark fields
@@ -188,6 +192,7 @@ values are merged with per section values.
 | | `type` | `exitcode-stdio-1.0` | |
 | `main` | `main-is` | | |
 | `other-modules` | · | All modules in `source-dirs` less `main` less any modules mentioned in `when` | |
+| `generated-other-modules` | | | Added to `other-modules` and `autogen-modules`. Since `0.22.0`.
 | | `default-language` | `Haskell2010` | |
 
 #### <a name="flags"></a>Flags
--- a/src/Hpack/Run.hs
+++ b/src/Hpack/Run.hs
@@ -132,7 +132,7 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
     cabalVersion = (">= " ++) . showVersion <$> maximum [
         Just (makeVersion [1,10])
       , packageCabalVersion
-      , packageLibrary >>= libraryCabalVersion . sectionData
+      , packageLibrary >>= libraryCabalVersion
       , internalLibsCabalVersion packageInternalLibraries
       , executablesCabalVersion packageExecutables
       , executablesCabalVersion packageTests
@@ -146,16 +146,16 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
         , makeVersion [1,18] <$ guard (not (null packageExtraDocFiles))
         ]
 
-      libraryCabalVersion :: Library -> Maybe Version
-      libraryCabalVersion Library{..} = maximum [
+      libraryCabalVersion :: Section Library -> Maybe Version
+      libraryCabalVersion sect = maximum [
           makeVersion [1,22] <$ guard hasReexportedModules
         , makeVersion [2,0]  <$ guard hasSignatures
         , makeVersion [2,0] <$ guard hasGeneratedModules
         ]
         where
-          hasReexportedModules = (not . null) libraryReexportedModules
-          hasSignatures = (not . null) librarySignatures
-          hasGeneratedModules = (not . null) libraryGeneratedModules
+          hasReexportedModules = any (not . null . libraryReexportedModules) sect
+          hasSignatures = any (not . null . librarySignatures) sect
+          hasGeneratedModules = any (not . null . libraryGeneratedModules) sect
 
       internalLibsCabalVersion :: Map String (Section Library) -> Maybe Version
       internalLibsCabalVersion internalLibraries = makeVersion [2,0] <$ guard (not (Map.null internalLibraries))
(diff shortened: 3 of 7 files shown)


quasicomputational:
I'll rebase this onto master anyway to keep the delta small.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
   - Accept section-specific fields in conditionals (see #175, thanks to Scott
     Fleischman)
   - New section: `internal-libraries`, for Cabal 2's internal libraries (see #200).
+  - `generated-exposed-modules` and `generated-other-modules`, for populating
+    the `autogen-modules` field (#207).
```
sol: This need to be moved to the top of the file now.

README.md
```diff
@@ -101,7 +101,9 @@ These fields are merged with all library, executable, test, and benchmark compon
 | --- | --- | --- | --- |
 | `exposed` | · | | |
 | `exposed-modules` | · | All modules in `source-dirs` less `other-modules` less any modules mentioned in `when` | |
+| `generated-exposed-modules` | | `[]` | Added to `exposed-modules` and `autogen-modules`. Since `0.21.0`.
```
sol: - I would leave default empty instead of saying `[]`.  This is the convention used in the README so far.
- I renamed the last column from `Notes` to `Default inside conditional`, so your content does not fit that column anymore.  What I would do is add `exposed-modules and autogen-modules` to the second column (`Cabal`).  If we want to specify `since`, we may want to have a column for that as we have for `Top-level fields`.

sol: Regarding `since`, this will tentatively go into `0.22.0`.

test/EndToEndSpec.hs
```diff
@@ -555,6 +729,53 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+              hs-source-dirs:
+                  src
+              other-modules:
+                  Baz
+                  XYZ
+              autogen-modules:
+                  XYZ
+              default-language: Haskell2010
+          |]
+
```
sol: Can you try to turn some of the acceptance tests into unit tests for e.g. `determineModules`.

The general direction I want to go with the tests is: Get rid of the integration tests in `ConfigSpec` and `RunSpec` and have acceptance tests in `EndToEndSpec` and unit tests instead.

test/EndToEndSpec.hs
```diff
@@ -555,6 +729,53 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+              hs-source-dirs:
+                  src
+              other-modules:
+                  Foo
+                  Paths_foo
+                  XYZ
+              autogen-modules:
+                  XYZ
+              default-language: Haskell2010
+          |]
```
sol: I think this can be
```
|] `shouldRenderTo` executable "foo" ([i|
main-is: Main.hs
hs-source-dirs:
    src
other-modules:
    Foo
    Paths_foo
    XYZ
autogen-modules:
    XYZ
|]) {packageCabalVersion = ">= 2.0"}
```
instead.

test/Hpack/RunSpec.hs
```diff
@@ -245,6 +259,24 @@ spec = do
...
+          , "executable foo"
+          , "  main-is: Main.hs"
+          , "  other-modules:"
+          , "      ABC"
+          , "  autogen-modules:"
+          , "      ABC"
+          , "  default-language: Haskell2010"
+          ]
+
+
```
sol: I would ditch this two tests.  Either we have it covered by the acceptance tests or have a function

```haskell
requiredCabalVersion :: Package -> Version
```

and unit tests for that function.

README.md
quasicomputational: The columns only seem to have been changed in the 'Library' section. Is it deliberate that it's still 'Notes' for executables, tests and benchmarks, as well as the common fields? Also, I notice that only `other-modules` has (documented) behaviour; it possibly makes more sense to go back to 'Notes'?

sol: Yes, works for me.

quasicomputational:
I've addressed most of those issues. The one I didn't get to was replacing the EndToEnd.hs tests with unit tests for determineModules: I added the latter, but what kept me from deleting the former is that, otherwise, some parts of the code aren't going to be exercised.

test/Hpack/ConfigSpec.hs
```diff
@@ -72,28 +72,43 @@ spec = do
...
+    determineModules ["Paths_foo"] [] Nothing ["ABC"] Nothing ["XYZ"] `shouldBe` (["ABC"], ["Paths_foo", "XYZ"], ["XYZ", "ABC"])
+
+  it "includes all generated modules when given explicit exposed-modules" $
+    determineModules ["Paths_foo"] [] ["Foo"] ["ABC"] Nothing ["XYZ"] `shouldBe` (["Foo", "ABC"], ["Paths_foo", "XYZ"], ["XYZ", "ABC"])
+
+  it "includes all generated modules when given explicit exposed-modules and other-modules" $
+    determineModules ["Paths_foo"] [] ["Foo"] ["ABC"] ["Internal"] ["XYZ"] `shouldBe` (["Foo", "ABC"], ["Internal", "XYZ"], ["XYZ", "ABC"])
+
+  it "includes all generated modules when given explicit other-modules" $
+    determineModules ["Paths_foo"] [] Nothing ["ABC"] ["Internal"] ["XYZ"] `shouldBe` (["ABC"], ["Internal", "XYZ"], ["XYZ", "ABC"])
```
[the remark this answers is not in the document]
sol: The readability of this tests is not that great.  It's not clear what we pass to `determineModules` without looking at it's implementation.  This wasn't great before, but with the number of arguments increased it's even worse now.

At the very least you may want to add local bindings that name the things that you pass to `determineModules`.  An other option could be to add a dedicated record type for the arguments, effectively simulating keyword arguments.

The tests are not extensive, e.g. the following acceptance test fails:
```haskell
touch "src/XYZ.hs"
[i|
library:
  source-dirs: src
  generated-exposed-modules: XYZ
|] `shouldRenderTo` (library [i|
hs-source-dirs:
    src
exposed-modules:
    XYZ
other-modules:
    Paths_foo
autogen-modules:
    XYZ
|]) {packageCabalVersion = ">= 2.0"}
```

This is a non-trivial feature.  Readability is crucial, for both the implementation and the tests, for both the sake of future maintainability and for making sure that this does the right thing in the first place.

 1. Don't be shy to use `context` to structure the tests
 2. Make sure that every test is minimal and tests exactly one thing

CHANGELOG.md
```diff
@@ -13,7 +16,6 @@
 ## Changes in 0.21.0
   - Accept section-specific fields in conditionals (see #175, thanks to Scott
     Fleischman)
-  - New section: `internal-libraries`, for Cabal 2's internal libraries (see #200).
```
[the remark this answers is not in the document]
quasicomputational: Oops, yes.

src/Hpack/Run.hs
```diff
@@ -142,17 +145,28 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
         | otherwise = Nothing
 
       libraryCabalVersion :: Library -> Maybe Version
-      libraryCabalVersion Library{..} = maximum [
-          makeVersion [1,22] <$ guard hasReexportedModules
+      libraryCabalVersion Library{..} = maximum
+        [ makeVersion [1,22] <$ guard hasReexportedModules
```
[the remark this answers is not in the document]
quasicomputational: Merging is hard :(

test/Hpack/ConfigSpec.hs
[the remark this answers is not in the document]
quasicomputational: That test's testing what I'd call a user error situation: cabal has a directory specifically to plonk autogenerated files in (`Distribution.Simple.BuildPaths.autogenComponentModulesDir`). Is it worth detecting and warning about this situation?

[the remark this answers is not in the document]
sol: Do you know of any packages that actually do this?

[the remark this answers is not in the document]
sol: > cabal has a directory specifically to plonk autogenerated files in (`Distribution.Simple.BuildPaths.autogenComponentModulesDir`)

I'm not saying that this won't work or that it is a bad idea.  It just wast not apparent to me.

Quoting what @ezyang said on #hackage:

> think of autogen-modules as a "bit" you can flip on an exposed/other modules. It's primarily for helping sdist out

> autogenComponentModulesDir refers to where Cabal puts autogenerated modules *it* generates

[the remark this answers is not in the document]
sol: Looking at the tests case for `autogen-modules`

https://github.com/haskell/cabal/commit/c0108673d34c5f1727994b421a2aabe407901373#diff-6735ab8f37906726ada75a2f33fa2fceR25

doesn't reveal where `MyHelperModule` is supposed to be located neither.

My naive assumption is that modules listed in `autogen-modules` can be in any of the source directories (and possibly in `Distribution.Simple.BuildPaths.autogenComponentModulesDir` too).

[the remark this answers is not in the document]
quasicomputational: OK, so this needs fixing (and testing the fixing). I'll get on that.

quasicomputational:
I think I've hit all the comments. `determineModules` still has an ugly signature but I dunno if that's within the scope of this PR. I also tried to use `ModulesSpecification` as a field of `LibrarySection` and `ExecutableSection`, but that messed with the generic JSON magic so I backed off from that.

quasicomputational:
Oh, Travis is red because of -Werror, reasonably enough. What do I do about that? Export `ModuleSpecification` and its field accessors unconditionally, even though it's only needed for tests?

sol:
> Oh, Travis is red because of -Werror, reasonably enough. What do I do about that? Export ModuleSpecification and its field accessors unconditionally, even though it's only needed for tests?

Prefixing a variable name with an underscore prevents this warning.  This also work for record fields.

src/Hpack/Run.hs
```diff
@@ -145,14 +148,25 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
       libraryCabalVersion Library{..} = maximum [
           makeVersion [1,22] <$ guard hasReexportedModules
         , makeVersion [2,0]  <$ guard hasSignatures
+        , makeVersion [2,0] <$ guard hasGeneratedModules
         ]
         where
           hasReexportedModules = (not . null) libraryReexportedModules
           hasSignatures = (not . null) librarySignatures
+          hasGeneratedModules = (not . null) libraryGeneratedModules
```
sol: This does not take conditional into account.

src/Hpack/Run.hs
```diff
@@ -145,14 +148,25 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
...
       internalLibsCabalVersion internalLibraries = makeVersion [2,0] <$ guard (not (Map.null internalLibraries))
 
+      executablesCabalVersion :: Map String (Section Executable) -> Maybe Version
+      executablesCabalVersion = foldr max Nothing . map executableCabalVersion . Map.elems
+
+      executableCabalVersion :: Section Executable -> Maybe Version
+      executableCabalVersion sect = makeVersion [2,0] <$ guard (executableHasGeneratedModules sect)
+
+      executableHasGeneratedModules :: Section Executable -> Bool
+      executableHasGeneratedModules = not . null . executableGeneratedModules . sectionData
```
sol: I think this does neither.

sol:
An other failing test case:
```haskell
      it "includes all generated modules" $ do
        touch "src/ABC.hs"
        [i|
        library:
          source-dirs: src
          when:
            condition: foo
            generated-exposed-modules: ABC
            generated-other-modules: XYZ
        |] `shouldRenderTo` (library [i|
        hs-source-dirs:
            src
        if foo
          exposed-modules:
              ABC
          other-modules:
              XYZ
          autogen-modules:
              XYZ
              ABC
        other-modules:
            Paths_foo
        |]) -- {packageCabalVersion = ">= 2.0"}
```

quasicomputational:
It looks like the neighbouring code is wrong as well: it won't set the right cabal-version if there's a signature or reexport inside a conditional. I'll go ahead and test/fix those while I'm at it.

quasicomputational:
Seems like there's a bug somewhere else, with `reexported-modules` getting lost inside conditionals. I'll investigate and see if I can turn it green.

quasicomputational:
Never mind. I just forgot how to keyboard.

sol:
I squashed everything but 
https://github.com/sol/hpack/pull/227/commits/74e6826a97e92ca72d6f3201fc81758ff0428ff1
into a single commit, rebased and pushed to `master`.  I've applied some refactoring on top of that.

sol:
Everything on master now. @quasicomputational thanks a lot and sorry for the delay!

PR #235, opened 2017-12-12
Allow dependency constraints to be numbers (close #234)

### LINKED ISSUE #234: Allow dependency constraints to be numbers
You can specify dependencies as a hash where the keys are package names and the values are either nothing, strings, or hashes. It would be nice if the values could be numbers too. For example:

``` yaml
dependencies:
  # This works fine because `0.5.7.1` is parsed as a string.
  http-client: 0.5.7.1
  # This doesn't work because `0.11` is parsed as a number.
  http-types: 0.11
```

This is the exact error that I see:

```
.../package.yaml: Error in $.dependencies['http-types']: expected Null, Object, or String, encountered Number
```

sol:


test/Hpack/DependencySpec.hs
```diff
@@ -109,8 +109,34 @@ spec = do
...
+            [i|
+              hpack: 0.11
+            |] `parsesAs` Right [("hpack", VersionRange "==0.11")]
+
+          it "accepts 0.110" $ do
+            [i|
+              hpack: 0.110
+            |] `parsesAs` Right [("hpack", VersionRange "==0.110")]
+
+          it "accepts 1e2" $ do
```
tfausak: 😆

test/Hpack/DependencySpec.hs
```diff
@@ -109,8 +109,34 @@ spec = do
...
+            [i|
+              hpack: 1e2
+            |] `parsesAs` Right [("hpack", VersionRange "==100")]
```
sol: @tfausak we could choose to reject `1e2` as the difference between `1e2` and `100` is observable.  However, the difference between `0.01` and `1e-2` is not observable.

tfausak: I think it's fine to allow `1e2`. It's not a valid version number, so there's no potential for confusion. It's kind of weird, but 🤷‍♂️

sol: Agree.

sol:
On Hackage.

Issue #237, opened 2017-12-12
Escaping of ghc-options with spaces


snoyberg:
Consider the following cabal file:

```
name:           foo
version:        0.1.0.0
build-type:     Simple
cabal-version:  >= 1.10

executable foo
  main-is: Main.hs
  build-depends: base
  default-language: Haskell2010
  ghc-options: -threaded -O2 -rtsopts "-with-rtsopts=-N -T"
```

Using `hpack-convert`, I end up with:

```yaml
name: foo
version: '0.1.0.0'
license: UnspecifiedLicense
ghc-options:
- -threaded
- -O2
- -rtsopts
- -with-rtsopts=-N -T
dependencies:
- base
executables:
  foo:
    main: Main.hs
```

Which results in the following in the generated cabal file:

```
  ghc-options: -threaded -O2 -rtsopts -with-rtsopts=-N -T
```

Unfortunately, this does not build:

```
ghc: unrecognised flag: -T
did you mean one of:
  -F
  -v
  -j
```

I can work around this by putting the following in my package.yaml instead of the list of options:

```
ghc-options: -threaded -O2 -rtsopts "-with-rtsopts=-N -T"
```

However, it would be nice to keep the list approach, and have `hpack` detect when there's a space in an option and automatically wrap it in quotes.

tfausak:
Sounds a lot like #182. I think fixing this in hpack (#238) makes more sense than requiring people to muck around with YAML syntax and GHC argument parsing.

sol:
@snoyberg thanks for opening this issue.

> I think fixing this in hpack (#238) makes more sense than requiring people to muck around with YAML syntax and GHC argument parsing.

Yes, I agree.  If there is a way to address this in hpack I'm all for it.  However, just quoting everything may do more harm than good.  So let's first collect all the facts.

Specifically:

1. Is `-with-rtsopts` the only affected GHC option?

1. Currently, the following is valid in `package.yaml`:
   ```yaml
   ghc-options: -Wall -Werror
   ```
   If we quote all options that contain spaces the above will become invalid.

1. We could consider to only quote options that were given as a YAML list, e.g.
   ```yaml
   ghc-options:
     - -with-rtsopts=-N -T
   ```
   we would rendered as `ghc-options: "-with-rtsopts=-N -T", while
   ```yaml
   ghc-options: -Wall -Werror
   ```
   we would continue to render as `ghc-options: -Wall -Werror`.

[...]

PR #238, opened 2017-12-12
Add quotes around values with spaces #237

snoyberg:
I'm not convinced this is the best way to solve this, but wanted to open the PR to get feedback. Three potential issues I can see:

* I'm using a simple blacklist of space characters; should we instead whitelist characters which do not require quoting?
* Do we need to also add logic for escaping values. For example, what should we do if one of the options itself contains a quote inside? I'm not familiar with the cabal format escaping rules here
* Is it OK to apply this logic to all examples of `WordList`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## next (0.21.2)
   - New top-level field: `extra-doc-files`.
+  - Add quotes around values with spaces in them (see #237)
 
 ## Changes in 0.21.1
   - Allow dependency constraints to be numbers (see #234)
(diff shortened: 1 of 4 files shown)


sol:
@snoyberg Hey, thanks for looking into this!  My main concern is that this would break existing package specifications, e.g.

```yaml
ghc-options: -Wall -Werror
```
is currently valid, but won't be valid with this change.  Let's try to hash thing out in #237.

sol:
Per the current discussion of #237 this approach is not viable.  Closing.

Issue #239, opened 2017-12-15
Infer signatures


sol:
We want to default the value of `signatures` based on the presence of any `.hsig`-files.

PR #240, opened 2017-12-15
Implement Infer signatures

waddlaw:
@sol Sorry, I'm late.

Ref to [#239](https://github.com/sol/hpack/issues/239).

How to determine signatures precedence.

1. specify `signature` field in `package.yaml`
2. lookup `.hsig` files if not specify `signature` in `package.yaml`
3. Nothing if not exist `.hsig` file.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -40,6 +40,7 @@ module Hpack.Config (
 , Empty(..)
 , getModules
 , pathsModuleFromPackageName
+, getSignatures
 , determineModules
 , BuildType(..)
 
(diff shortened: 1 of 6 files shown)


test/Hpack/ConfigSpec.hs
```diff
@@ -157,6 +158,14 @@ spec = do
         touch (dir </> "Setup.hs")
         getModules dir  "./." `shouldReturn` ["Foo"]
 
+  describe "getSignatures" $ around withTempDirectory $ do
+    it "returns signatures string in directory" $ \dir -> do
+      touch (dir </> "./Foo.hs")
+      touch (dir </> "./Bar/Baz.hs")
+      touch (dir </> "./Test.hsig")
+      touch (dir </> "./Test2.hsig")
+      getSignatures dir >>= (`shouldBe` Just (List ["Test", "Test2"]))
```
sol: This rejects hierarchical signatures, e.g. `Foo.Bar`.

Are hierarchical signatures allowed? cc @ezyang

src/Hpack/Config.hs
```diff
@@ -803,39 +804,40 @@ inferModules ::
...
   -> IO (Section b)
 inferModules dir packageName_ getMentionedModules getInferredModules fromData fromConditionals = traverseSectionAndConditionals
   (fromConfigSection fromData [pathsModuleFromPackageName packageName_])
-  (fromConfigSection (\ [] -> fromConditionals) [])
+  (fromConfigSection (\ _ [] -> fromConditionals) [])
   []
   where
     fromConfigSection fromConfig pathsModule_ outerModules sect@Section{sectionData = conf} = do
       modules <- listModules dir sect
+      inferSignatures <- getSignatures dir
```
sol: In addition, this looks for signatures next to `package.yaml` (in contrast to `source-dirs`).  My naive assumption would be that signatures have to be inside `source-dirs` too.  But then again, I'm not sure, as documentation is missing and the examples I saw don't specify `hs-source-dirs`.

@ezyang can you clarify?

test/Hpack/ConfigSpec.hs
ezyang: Yes, hierarchical signatures are allowed; the set of valid module names for signatures coincides with the set of valid module names.

waddlaw: I understand. For example;

```
./Bar/Baz.hs
./Foo.hsig
./Foo/Bar.sig
```

Should be listed `Foo`, `Foo.Bar` in signatures?
I will be rewrite some code if correct.

waddlaw:
@sol 
What about the problems (`hierarchical signatures` and `source-dirs`) ?

Issue #244, opened 2017-12-24
Defaults: directory traversal attack


quasicomputational:
I do not think that this implementation of defaults is secure against directory traversal. Consider this package.yaml:
```
defaults:
  github: example/evil
  ref: master
  path: ../../../../.ssh/id_rsa
```
That's going to fail because an SSH private key (probably?) won't parse as YAML, but I am sure someone cleverer than me can lever this into a workable attack. There are two different aspects to this:
1. The ability to drop random files anywhere user-writable when hpack is run (very possible, fairly dangerous)
2. The ability to read any user-readable file and potentially include parts of them in the generated cabal file (harder to exploit but far from impossible, and very very very dangerous).

sol:
@quasicomputational Hey, thank you for opening this issue.

Even if I'm going to try to challenge the attack scenarios, I'm still more than willing to address this, so that future code changes or things we miss won't lead to harmful behavior in the future.

> The ability to drop random files anywhere user-writable when hpack is run (very possible, fairly dangerous)

Good point.  Not a useful feature, we should address this.

> The  ability to read any user-readable file and potentially include parts of  them in the generated cabal file (harder to exploit but far from  impossible, and very very very dangerous).

This is already possible with includes, e.g. from `README.md`

```yaml
tests:
  hlint: !include "../common/hlint.yaml"
```

further more, I was planning to implement support for locally defined defaults, e.g.

```yaml
defaults:
  path: ../common.yaml
```

Basically, what it means is "if you can run `hpack` (or `stack` for that matter) then you can potentially read all user-readable files".  Is this an issue?  If yes, how to address?

This that we can do is:
 - Reject includes in defaults files
 - At the point we both
    - have locally defined defaults
    - accept defaults recursively

   we can reject locally defined defaults in defaults files.

quasicomputational:
Fair point on the include mechanism already providing a way to get at any user-readable files. With that in mind, being able to directory-traverse your way to any file via defaults isn't making things worse. Further, I agree local defaults with relative paths are a very desirable feature, so maybe (2) isn't a problem.

My thinking is that, either way, hpack knows that all legitimate accesses to the defaults cache, read or write, are going to be under `$XDG_HOME/hpack/defaults`. On Linux (and I think FreeBSD and maybe OpenBSD now too), you can use `openat` to get that guarantee, but Windows would obviously need a different solution. Alternatively, to stop (1), hashing the URL being fetched and using that as the filename is also an option.

sol:
[...]

PR #245, opened 2017-12-26
Sanitize defaults

sol:
Addresses #244.

### DIFF
--- a/src/Hpack/Syntax.hs
+++ b/src/Hpack/Syntax.hs
@@ -1,9 +1,15 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE RecordWildCards #-}
 module Hpack.Syntax (
   Defaults(..)
+#ifdef TEST
+, isValidUser
+, isValidRepo
+#endif
 ) where
 
+import           Data.List
 import           Data.Maybe
 import           Data.Data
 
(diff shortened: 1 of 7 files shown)


src/Hpack/Syntax.hs
```diff
@@ -1,40 +1,110 @@
...
+        | isValidRef ref = return (Ref ref)
+        | otherwise = fail ("invalid reference " ++ show ref)
+
+data Path = Path {unPath :: [FilePath]}
+
+instance FromJSON Path where
+  parseJSON v = parseJSON v >>= parsePath
+    where
+      parsePath path
+        | '\\' `elem` path = fail ("rejecting '\\' in " ++ show path ++ ", please use '/' to separate path components")
```
quasicomputational: Backslash is a perfectly cromulent path character in POSIX, though I guess there's not too much harm in subsetting for Windows compatibility.

src/Hpack/Syntax.hs
```diff
@@ -1,40 +1,110 @@
...
+        | "/" `elem` p = fail ("rejecting absolute path " ++ show path)
+        | ".." `elem` p = fail ("rejecting \"..\" in " ++ show path)
+        | otherwise = return (Path p)
```
quasicomputational: Are we sure there are no weird Windows corner cases here? I'd also want to double check that `~` won't get expanded to `$HOME`.

src/Hpack/Syntax.hs
quasicomputational: Forgot to say that, if we do subset POSIX paths for compatibility, that should be documented.

src/Hpack/Syntax.hs
sol: > Are we sure there are no weird Windows corner cases here?

Apparently we need to exclude `:`.

> I'd also want to double check that ~ won't get expanded to $HOME.

It won't, `~` is expanded by the shell.

PR #247, opened 2018-01-07
Set cabal-version to >= 1.18 when using extra-doc-files.

quasicomputational:
Managed to miss this until I was investigating something else and cabal check started complaining at me!


sol:
Thanks a lot!

PR #250, opened 2018-01-08
Defaults from local files.

### LINKED ISSUE #248: Allow pulling defaults from local files
This would fit the monorepo design better, as well as being an escape route for anyone who doesn't want to rely on GitHub.

quasicomputational:
Closes #248.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
   - Remove Git conflict markers before checking the hash of any existing
     `.cabal` files (equivalent to `git checkout --ours`).  This allows you to
     regenerate the `.cabal` on conflicts when rebasing without passing `-f`.
+  - Allow local files to be used as defaults (#248)
 
 ## Changes in 0.25.0
   - Keep non-existing literal files on glob expansion (see #101)
--- a/README.md
+++ b/README.md
@@ -83,10 +83,11 @@ on the features that are used.
 #### <a name="defaults"></a>Defaults
 
 Hpack allows the inclusion of [common fields](#common-fields) from a file on
-GitHub.
+GitHub or a local file.
 
 To use this feature a user must specify a GitHub repository, Git reference and
-a path to a file within that repository.
+a path to a file within that repository; alternatively, a path to the local
+file must be given.
 
 Example:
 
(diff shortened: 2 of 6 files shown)


sol:
@quasicomputational thanks for opening this PR and sorry for the delay.  I did a major refactoring on how to determine unrecognized fields.  The good news is that it now does not require any special attention in the AST anymore.  But as a result this PR requires rebasing.

Regarding concrete syntax, I would want this to be called `path` instead of `local-path`.  That is, if `path` is specified without `github` then it's a local path.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
   - `extra-doc-files` requires setting `cabal-version` to at least
     1.18; this is now done properly.
   - Accept bool for `condition` (see #230)
+  - Allow local files to be used as defaults (#248)
```
sol: Please also add `(fixes #248)` to the commit message, so that GitHub will close the issue automatically once this goes on `master`.

src/Hpack/Config.hs
```diff
@@ -452,7 +452,7 @@ data PackageConfig capture cSources jsSources = PackageConfig {
 , packageConfigExecutables :: Maybe (Map String (SectionConfig capture cSources jsSources ExecutableSection))
 , packageConfigTests :: Maybe (Map String (SectionConfig capture cSources jsSources ExecutableSection))
 , packageConfigBenchmarks :: Maybe (Map String (SectionConfig capture cSources jsSources ExecutableSection))
-, packageConfigDefaults :: Maybe (capture Defaults)
+, packageConfigDefaults :: Maybe (Defaults capture)
```
sol: `capture` won't be necessary anymore

quasicomputational:
> Regarding concrete syntax, I would want this to be called path instead of local-path. That is, if path is specified without github then it's a local path.

That was what I initially went for, but then I got worried about forwards compatibility. Suppose that this gets merged for hpack 0.24 and then a way to grab defaults from Bitbucket gets added in hpack 0.25. It'd look quite like the Github version:

```
defaults:
  bitbucket: foo/bar
  path: defaults.yaml
```

But hpack 0.24 will interpret that as a local path (and complain about the unknown field), and that might lead to surprising behaviour if you have a `defaults.yaml` in your project for some reason.

Hence, I thought that it'd be better to say 'get the defaults from this local file' with `local-path`, since that will unambiguously identify where the defaults ought to come from, even if there are some random other fields added in the future.

I agree that `path` instead of `local-path` is nicer. Is this scenario too far-fetched?

sol:
> Is this scenario too far-fetched?

I think it's a valid concern.  But as we have the warning I think it's still fine to go with `path`.

I guess ptions (1) and (2) in #261 would thwart that concern.

sol:
@quasicomputational after giving this more thought, how about we use `local` for local defaults?

### Examples

GitHub:

```yaml
defaults:
  github: owner/repo
  ref: ..
  path: ..
```

Local:

```yaml
defaults:
  local: defaults.yaml
```

quasicomputational:
OK! I'm going to rebase and get back to this, but possibly not for a few days.

quasicomputational:
:thumbsup: No objections, that's all sensible. I probably do over-test, especially black-box whole-system tests! I got bit by a test-first werewolf years ago and I haven't found the cure yet; later I got burnt by white-box tests ossifying interfaces prematurely.

sol:
> I got bit by a test-first werewolf years ago

Thats not so bad!  Much better than being ignorant to testing!  For me proper testing is still an evolving discipline were we keep experimenting and learning.

> burnt by white-box tests ossifying interfaces prematurely

From my experience integration tests (or unit tests that test too much) are the main culprit here.  Unit tests (provided they are scoped properly) are easier to manage, easier to refactor together with the code they test, and easier to throw away when that code becomes obsolete.

sol:
@quasicomputational on Hackage as `hpack-0.26.0`

PR #257, opened 2018-01-21
Refactoring for generated-*-modules

sol:


src/Hpack/Config.hs
```diff
@@ -1037,36 +1037,48 @@ toLibrary dir name globalOptions =
...
         (exposedModules, otherModules, generatedModules) =
-          determineModules pathsModule inferableModules (ModuleSpecification librarySectionExposedModules librarySectionGeneratedExposedModules) (ModuleSpecification librarySectionOtherModules librarySectionGeneratedOtherModules)
+          determineModules DetermineModules {
+              _pathsModule = pathsModule
+            , _inferableModules = inferableModules
+            , _exposedModules = librarySectionExposedModules
+            , _generatedExposedModules = librarySectionGeneratedExposedModules
+            , _otherModules = librarySectionOtherModules
+            , _generatedOtherModules = librarySectionGeneratedOtherModules
+          }
```
sol: This is what I meant with simulating keyword arguments, not necessarily the right thing to do, just to make the point.

src/Hpack/Config.hs
```diff
@@ -996,8 +996,8 @@ traverseSectionAndConditionals fData fConditionals acc0 sect@Section{..} = do
     traverseConditionals = traverse . traverse . traverseSectionAndConditionals fConditionals fConditionals
 
 getMentionedLibraryModules :: LibrarySection -> [String]
-getMentionedLibraryModules LibrarySection{..} =
-  fromMaybeList librarySectionExposedModules ++ fromMaybeList librarySectionOtherModules
+getMentionedLibraryModules (LibrarySection _ exposedModules generatedExposedModules otherModules generatedOtherModules _ _)
+  = fromMaybeList (exposedModules <> generatedExposedModules <> otherModules <> generatedOtherModules)
```
sol: Generated modules are never inferred because they are explicitly mentioned.  This fact was duplicated in several places before.

test/Hpack/ConfigSpec.hs
```diff
@@ -71,41 +82,26 @@ spec = do
...
+                _exposedModules = exposedModules
+              , _generatedExposedModules = generatedExposedModules
+              , _otherModules = otherModules
+              , _generatedOtherModules = generatedOtherModules
+              }
+            determineModules d `shouldBe` (
+                fromMaybeList exposedModules <> fromMaybeList generatedExposedModules
+              , maybe ["Paths_foo"] fromList otherModules <> fromMaybeList generatedOtherModules
+              , fromMaybeList (generatedExposedModules <> generatedOtherModules)
+              )
```
sol: In terms of code penetration, this test does everything the previous tests did.

We didn't test anything related to inferable modules though.

src/Hpack/Config.hs
sol: @quasicomputational

quasicomputational: :thumbsup:

Definitely cleaner this way than trying to exploit the parallel between exposed and other modules.

test/Hpack/ConfigSpec.hs
quasicomputational: My only worry here is that this is quite white-box, but I guess the end-to-end tests will exercise determineModules in a black-box way.

PR #259, opened 2018-02-01
Get package library/test/etc. dependencies separately

barischrooneyj:
In another project I want to programmatically add dependencies to a package, starting with adding to a package library's dependencies, so I have introduced a function I need:

``` Haskell
packageLibraryDependencies :: Package -> Dependencies
```

I also added the respective functions for executables, tests and benchmarks.

This also allowed the function `packageDependencies` to be written in terms of these functions.

***

I also have a few questions:

1. in `packageDependencies` each of the `Dependencies` are `mconcat`ed together now and then finally converted `toList`, instead of converting to lists first and using list concatenation. (I think this makes more semantic sense, i.e. we're combining Dependencies so let's use the function that exists for that purpose). My question is, now using `mconcat`, can `nub` be removed, I think yes but I'm not sure.

2. I think for a consistent API any function called `fooDependencies` should return a `Dependencies`. So I would change `packageDependencies` to be simply:

``` Haskell
packageDependencies :: Package -> Dependencies
packageDependencies p = mconcat [
    packageExecutableDependencies p
  , packageBenchmarkDependencies p
  , packageTestDependencies p
  , packageLibraryDependencies p
  ]
```

and then introduce `packageDependenciesSorted` as:

``` Haskell
packageDependenciesSorted :: Package -> [(String, DependencyVersion)]
packageDependenciesSorted = nub . sortBy (comparing (lexicographically . fst)) . Exts.toList . packageDependencies
```

This would be an API breaking change, thus would be in a new PR and would have to wait for next major version bump.

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -33,6 +33,10 @@ module Hpack.Config (
 
 , renamePackage
 , packageDependencies
+, packageExecutableDependencies
+, packageTestDependencies
+, packageBenchmarkDependencies
+, packageLibraryDependencies
 , package
 , section
 , Package(..)
@@ -91,6 +95,7 @@ import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Writer
 import           Control.Monad.Trans.Except
 import           Control.Monad.IO.Class
+import qualified GHC.Exts as Exts
 
 import           Data.Aeson.Config.Types
 import           Data.Aeson.Config.FromValue hiding (decodeValue)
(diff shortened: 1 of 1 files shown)


sol:
@barischrooneyj Hi, thanks for using `hpack` and for taking the time to contribute!

> 1. in `packageDependencies` each of the `Dependencies` are `mconcat`ed together now and then finally converted `toList`, instead of converting to lists first and using list concatenation. (I think this makes more semantic sense, i.e. we're combining Dependencies so let's use the function that exists for that purpose). My question is, now using `mconcat`, can `nub` be removed, I think yes but I'm not sure.

At the point you are not sure it's probably time to write a test to verify your assumption ;)

I think concatenating on `Map` will not work as it potentially discards dependency versions, e.g. consider:

```yaml
library:
  dependencies:
    foo: null

executable:
  dependencies:
    foo: 0.1.0
```

> 2. I think for a consistent API any function called `fooDependencies` should return a `Dependencies`.

I never imagined that Hpack will be widely used as a library and zero thought went into API design. The stuff that is exposed is mainly motivated by what we need for `stack` and `tinc` integration.

That said, I'm happy to consider any changes that help with making Hpack more useful!

sol:
> This would be an API breaking change, thus would be in a new PR and would have to wait for next major version bump.

Most changes to Hpack break the API in some way (mostly by adding record filed to some type), so this is not a real concern alone.  But, `packageDependencies` is used by `tinc` and it needs access to the unfiltered list of dependencies to get all source dependencies (local and git), so that it can install/cache them correctly.

src/Hpack/Config.hs
```diff
@@ -129,14 +134,37 @@ renameDependencies old new sect@Section{..} = sect {sectionDependencies = (Depen
...
+-- | A list of a package's dependencies sorted lexicographically by name.
 packageDependencies :: Package -> [(String, DependencyVersion)]
-packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . fst)) $
-     (concatMap deps packageExecutables)
-  ++ (concatMap deps packageTests)
-  ++ (concatMap deps packageBenchmarks)
-  ++ maybe [] deps packageLibrary
-  where
-    deps xs = [(name, version) | (name, version) <- (Map.toList . unDependencies . sectionDependencies) xs]
+packageDependencies p = nub . sortBy (comparing (lexicographically . fst)) $ Exts.toList $ mconcat [
```
sol: As mentioned, `tinc` requires all dependencies, so `mconcat` on `Map` won't work.

sol:
Closing for inactivity.  @barischrooneyj please feel free to let me know in case you still want to do something here.

PR #260, opened 2018-02-02
Add verbatim as a Cabal escape hatch

sol:
This is meant to provide a temporary workaround for stuff that is not (yet) supported by Hpack.

Example usage:

```yaml
verbatim: |
  build-tool-depends:
      hspec-discover:hspec-discover == 2.*
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## next
   - Allow `version` be a numbers
   - Ignore fields that start with an underscore everywhere, not just globally
+  - Add support for `cabal` escape hatch
 
 ## Changes in 0.23.0
   - Add support for custom decoders to allow for alternate syntax (e.g. Dhall)
--- a/README.md
+++ b/README.md
@@ -281,6 +281,22 @@ Glob patterns are expanded according to the following rules:
    separators)
  - `?`, `*` and `**` do not match a `.` at the beginning of a file/directory
 
+### Passing things verbatim to Cabal
+
+Hpack is meant to be general and handle most common use cases.  However, in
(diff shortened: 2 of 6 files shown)


sol:
I would appreciate any feedback on this, esp. regarding the choice of `verbatim`.  Are there any better options?

mgsloan:
Can't think of any better options, seems good :+1: .  Escape hatches don't need to be particularly pretty.

sol:
Thanks guys!

vlatkoB:
If I specify top-level
```
verbatim: |
  data-dir:    data/
```
after `legal` part in package.yaml, the line

    data-dir:    data/
is appended to the very end of .cabal file, instead of inserted at its original position in `package.yml`.
GHC complains with
`Unable to parse cabal file .../app.cabal: FromString "Plain fields are not allowed in between stanzas: F 152 \"data-dir\" \"data/\"" (Just 152)`

sol:
@vlatkoB If you can provide a fully self-contained example `package.yaml` that reproduces the problem, then I can take a quick look later.

vlatkoB:
@sol: Using `hpack 0.26.0`
```
# package.yaml
name:        test
version:     '0.1.0.0'

verbatim: |
  data-dir:    data/

library:
  exposed-modules:
    - Test.hs
```
Generated .cabal:
```
# test.cabal
name:           test
version:        0.1.0.0
build-type:     Simple
cabal-version:  >= 1.10

library
  exposed-modules:
      Test.hs
  other-modules:
      Paths_test
  default-language: Haskell2010

data-dir:    data/
```

Issue #256, opened 2018-01-20
Release binaries for CI use?


quasicomputational:
For hpackified projects that can't/won't use stack (e.g., using features from more recent hpacks than the most recent stack release is built against), there's a need to run hpack as a separate step before the main build. On development machines, that's just an occasional chore (build hpack, stick it in ~/.local/bin, adjust build scripts and forget about it), but it's more painful on CI: building hpack takes upwards of 10 minutes on Travis. It's cacheable, so it's not entirely terrible, but it's one more papercut.

Checking in the generated cabal files is an option, but that has all the problems of having generated files in the VCS.

It'd be very handy to be able to grab a cutting-edge hpack binary from somewhere and use that directly. For example, hlint uses Github's 'releases' feature to provide binaries, and then [lens's Travis configuration](https://github.com/ekmett/lens/blob/master/.travis.yml#L84) grabs a binary and uses that.

This is very much a quality-of-life request and, if it's annoying, feel free to ignore it.

sol:
Yes, we absolutely want this!  For [`sol/tinc`](https://github.com/sol/tinc) I do releases to AWS S3 on every push to maser.

@quasicomputational Would you only want to upload tagged releases or every push to master?  If only tagged releases, then GitHub might be a great choice!

quasicomputational:
Just tagged releases would be wonderful! I'm okay with the idea that HEAD is unstable and should only be used directly in extremis, but it's a shame if we can't use 0.22's shinies easily just because stack is on a different release cadence.

sol:
Not sure when I'll have time to look into this.  A tested PR that just requires me to change the OAuth token would make this happen much faster and would be very much appreciated ;)

sol:
Let's see if this works out!

@quasicomputational do we need Windows binaries too?

sol:
We now have binaries for OSX and Linux at https://github.com/sol/hpack/releases/latest.  However, I can't figure out any URL to one of the latest binaries...

If anybody can figure it out, please let me know.  If it's not possible, then not sure how you would want to use on CI (assuming here that you always want to use the latest version).

tfausak:
As far as I know, there is no stable URL for the latest binary. You can use the API to find the latest release, then get the binary from that. For example: <https://www.starkandwayne.com/blog/how-to-download-the-latest-release-from-github/>.

sol:
That work too!

quasicomputational:
[...]

PR #267, opened 2018-02-05
Add get-hpack.sh (see #256)

sol:


get-hpack.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -o nounset
+set -o errexit
+
+dst="$HOME/.local/bin"
```
sol: ^ I assume this is already on the `$PATH` on travis?!

Issue #205, opened 2017-10-18
Feature request: add `cxx-sources`/`cxx-options` field


TerrorJack:
`Cabal` has added support for compiling C++ sources and added `cxx-sources`/`cxx-options` to the configuration (see [here](https://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-cxx-sources)). It would be nice for hpack to support this :)

sol:
PR please! ;)

PR #271, opened 2018-02-10
add cxx-sources/cxx-options fields

budevg:
solves #205

### DIFF
--- a/src/Hpack/Render.hs
+++ b/src/Hpack/Render.hs
@@ -148,6 +148,7 @@ renderPackageWith settings headerFieldsAlignment existingFieldOrder sectionsFiel
           makeVersion [1,22] <$ guard hasReexportedModules
         , makeVersion [2,0]  <$ guard hasSignatures
         , makeVersion [2,0] <$ guard hasGeneratedModules
+        , makeVersion [2,2] <$ guard (hasCxxParams sect)
         ]
         where
           hasReexportedModules = any (not . null . libraryReexportedModules) sect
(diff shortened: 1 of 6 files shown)


budevg:
Thanks for the review remarks ... adding commit with the changes

src/Hpack/Render.hs
```diff
@@ -127,6 +127,7 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
           makeVersion [1,22] <$ guard hasReexportedModules
         , makeVersion [2,0]  <$ guard hasSignatures
         , makeVersion [2,0] <$ guard hasGeneratedModules
+        , makeVersion [2,2] <$ guard (hasCxxParams sect)
```
sol: I think right now `internal-libraries` are not taken into consideration.  What we probably want to do is use `libraryCabalVersion` in `internalLibsCabalVersion` similar to what we already with `executableCabalVersion` and `executablesCabalVersion`.

src/Hpack/Render.hs
```diff
@@ -140,11 +141,21 @@ renderPackage settings alignment existingFieldOrder sectionsFieldOrder Package{.
...
+        ]
+
 
       executableHasGeneratedModules :: Section Executable -> Bool
       executableHasGeneratedModules = any (not . null . executableGeneratedModules)
 
+      hasCxxParams sect = or [
+          (not . null . sectionCxxOptions) sect
+        , (not . null . sectionCxxSources) sect
```
sol: Does this take effect for `cxx-*` in conditionals?  You may need to use `any` here.

test/EndToEndSpec.hs
```diff
@@ -434,6 +434,43 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+          [i|
+          library:
+            cxx-sources:
+              - foo.cc
+              - cxxbits/*.cc
+          |] `shouldRenderTo` (library_ [i|
+          cxx-sources:
+              cxxbits/bar.cc
+              foo.cc
+          |]) {packageCabalVersion = ">= 2.2"}
```
sol: My naive assumption is still that this last test is not required to drive the implementation, but I'm happy to be convinced otherwise ;)

budevg:
some more changes, rebase and squash

sol:
@budevg Hey! I was in bed with fever, sorry for the late reply!  This looks good now! Thanks a lot!

budevg:
Thanks ! Hope you fill better soon

PR #273, opened 2018-02-14
Resolve local paths to defaults relative to the file, not CWD.

### LINKED ISSUE #272: Local defaults resolve paths relative to CWD, not package.yaml
```
$ cat defaults.yaml 
ghc-options: A
$ cat foo/defaults.yaml 
ghc-options: B
$ cat foo/package.yaml 
name: foo
defaults:
  local: defaults.yaml

executable:
  main: Main.hs
$ hpack foo
generated foo/foo.cabal
$ cat foo/foo.cabal 
-- This file has been generated from package.yaml by hpack version 0.26.0.
--
-- see: https://github.com/sol/hpack
--
-- hash: 50483a81cfe49270d0609ad93e473197e7a19178bb79a3668b0e4fb06eaab4a4

name:           foo
version:        0.0.0
build-type:     Simple
cabal-version:  >= 1.10

executable foo
  main-is: Main.hs
  other-modules:
      Paths_foo
  ghc-options: A
  default-language: Haskell2010
$ cd foo
$ hpack
generated foo.cabal
$ cat foo.cabal 
-- This file has been generated from package.yaml by hpack version 0.26.0.
--
-- see: https://github.com/sol/hpack
--
-- hash: ac53f37dfeb39e79b5fc69c8040ceb5e13d164194f40506a34feaefc57af5fdf

name:           foo
version:        0.0.0
build-type:     Simple
cabal-version:  >= 1.10

executable foo
  main-is: Main.hs
  other-modules:
      Paths_foo
  ghc-options: B
  default-language: Haskell2010
```

Oops. Guess this is my mistake so I'll prepare a PR.

quasicomputational:
Closes #272.

### DIFF
--- a/src/Hpack/Defaults.hs
+++ b/src/Hpack/Defaults.hs
@@ -61,11 +61,10 @@ ensure userDataDir dir = \ case
       Found -> return (Right file)
       NotFound -> return (Left $ notFound url)
       Failed err -> return (Left err)
-  DefaultsLocal (Local file) -> do
-    let file' = dir </> file
-    doesFileExist file' >>= \ case
-      True -> return (Right file')
-      False -> return (Left $ notFound file')
+  DefaultsLocal (Local ((dir </>) -> file)) -> do
+    doesFileExist file >>= \ case
+      True -> return (Right file)
+      False -> return (Left $ notFound file)
   where
     notFound file = "Invalid value for \"defaults\"! File " ++ file ++ " does not exist!"
 
--- a/src/Hpack/Defaults.hs
+++ b/src/Hpack/Defaults.hs
@@ -51,20 +51,21 @@ get url file = do
 formatStatus :: Status -> String
 formatStatus (Status code message) = show code ++ " " ++ B.unpack message
 
-ensure :: FilePath -> Defaults -> IO (Either String FilePath)
-ensure dir = \ case
+ensure :: FilePath -> FilePath -> Defaults -> IO (Either String FilePath)
+ensure userDataDir dir = \ case
   DefaultsGithub defaults -> do
     let
       url = defaultsUrl defaults
-      file = defaultsCachePath dir defaults
+      file = defaultsCachePath userDataDir defaults
     ensureFile file url >>= \ case
       Found -> return (Right file)
       NotFound -> return (Left $ notFound url)
       Failed err -> return (Left err)
-  DefaultsLocal (Local file) ->
-    doesFileExist file >>= \ case
-      True -> return (Right file)
-      False -> return (Left $ notFound file)
+  DefaultsLocal (Local file) -> do
+    let file' = dir </> file
+    doesFileExist file' >>= \ case
+      True -> return (Right file')
+      False -> return (Left $ notFound file')
   where
     notFound file = "Invalid value for \"defaults\"! File " ++ file ++ " does not exist!"
 
--- a/test/HpackSpec.hs
+++ b/test/HpackSpec.hs
@@ -6,6 +6,8 @@ import           Prelude hiding (readFile)
 import qualified Prelude as Prelude
 
 import           Control.DeepSeq
+import           System.Directory (createDirectory)
+import           Test.Hspec (shouldContain)
 
 import           Hpack.Config
 import           Hpack.CabalFile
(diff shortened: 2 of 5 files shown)


quasicomputational:
Also, FTR, this matches the behaviour of `!include`.

quasicomputational:
Actually I think this patch must be wrong and the tests are passing by coincidence.

sol:
This is certainly a bug and we want to get this on Hackage soon.

> Actually I think this patch must be wrong and the tests are passing by coincidence.

Huh?  I can't see why.

quasicomputational:
The second test is actually not testing what it should be testing; all paths are being resolved relative to the original `package.yaml`, no matter where the `defaults.yaml` that we're looking at is. Having a look now and seeing if I can fix it...

sol:
> all paths are being resolved relative to the original `package.yaml`, no matter where the defaults.yaml that we're looking at is

Ah yes, you are right.

BTW, thew Windows build is broken.  I pushed a commit to master that should hopefully fix it.  Please rebase.

quasicomputational:
:thumbsup: Synchronicity!

src/Hpack/Defaults.hs
```diff
@@ -51,20 +51,21 @@ get url file = do
...
     ensureFile file url >>= \ case
       Found -> return (Right file)
       NotFound -> return (Left $ notFound url)
       Failed err -> return (Left err)
-  DefaultsLocal (Local file) ->
-    doesFileExist file >>= \ case
-      True -> return (Right file)
-      False -> return (Left $ notFound file)
+  DefaultsLocal (Local file) -> do
+    let file' = dir </> file
```
sol: I avoid using `'` in identifiers because admittedly, I use vim.  They don't work well with vim's `*`-search, and while vim allows you to adjust the identifier characters, adding `'` in turn won't play nice with `c2hs` which I used a lot in the past.  For this reason I tend to use `_` instead.

But in this particular case I would even eliminate the need for it by using view patterns.  That way we make sure that we can't possibly use the wrong `file` anywhere.

sol:
This looks good, except for the tests...

sol:
Regarding tests, I think half of the story can be covered by https://github.com/sol/hpack/commit/ab2fcf92cb6d7a27e5ec88f6d9f036c576a84ecf.  This is generally a good idea, as we never want to depend on the current working directory (also relevant for e.g. glob expansion).

What is left is that we adjust the path for recursive defaults.  With the current code structure it's hard to unit test this; we could have a second acceptance tests for it, but adapting the existing acceptance test is what I did https://github.com/sol/hpack/commit/0698800099c498d0c46c545f2dacaee906690432.

test/HpackSpec.hs
```diff
@@ -94,3 +96,64 @@ spec = do
...
+              , "  local: " ++ (defaultsDir </> "defaults.yaml")
+              , "executable:"
+              , "  main: Main.hs"
+              ]
+            _ <- hpackResult $ setTarget "package.yaml" defaultOptions
+            res <- readFile "foo.cabal"
+            res `shouldContain` "ghc-options: A"
+
+    context "with a relative path to local defaults" $ do
+      it "resolves relative to package.yaml, not CWD" $ do
```
sol: Covered by my two test cases.

test/HpackSpec.hs
```diff
@@ -94,3 +96,64 @@ spec = do
               old <- readFile file
               hpack `shouldReturn` outputUnchanged
               readFile file `shouldReturn` old
+
+    context "with an absolute path to local defaults" $ do
+      it "resolves the path correctly" $ do
+        withTempDirectory $ flip withCurrentDirectory $ do
```
sol: I'm dubious whether this tests any more than that given an absolute path `a`, `b </> a == a` for arbitrary `b`.  If we think we need to test something here, then I would prefer a unit test for `ensure`.

test/HpackSpec.hs
```diff
@@ -94,3 +96,64 @@ spec = do
...
+            , "executable:"
+            , "  main: Main.hs"
+            ]
+          _ <- hpackResult $ setTarget "foo/package.yaml" defaultOptions
+          fromOutside <- readFile "foo/foo.cabal"
+          _ <- withCurrentDirectory "foo" $ hpackResult defaultOptions
+          readFile "foo/foo.cabal" `shouldReturn` fromOutside
+
+      it "does not have false positives in cycle detection due to same-named default files in different directories" $ do
+        withTempDirectory $ flip withCurrentDirectory $ do
```
sol: Hmm, yes, that would be bad, but I think the code would actually get more complicated to implement this bug, say the simpler code gives us the correct behavior.

Usually I refrain from exhaustive testing and only write tests that are necessary to drive the implementation.

sol:
@quasicomputational please don't be offended by my harsh comments.  I think it's completely fine to write these tests in the first place to probe for possible bugs.  But then we either throw them away or turn them into unit tests.  Acceptance test driven development has scalability problems, if we now do exhaustive acceptance testing then that would even go one step further.

sol:
An last but not least, thanks for fixing this!  I'll try to get it on Hackage soon, probably as a minor release, I consider this a bug fix, the old behavior was hardly useful.

Issue #280, opened 2018-03-11
Top-level use of verbatim can result in invalid .cabal file


sol:
Steps to reproduce:

```yaml
# package.yaml
name:        test
version:     '0.1.0.0'

verbatim: |
  data-dir:    data/

library:
  exposed-modules:
    - Test.hs
```
Generated .cabal:
```
# test.cabal
name:           test
version:        0.1.0.0
build-type:     Simple
cabal-version:  >= 1.10

library
  exposed-modules:
      Test.hs
  other-modules:
      Paths_test
  default-language: Haskell2010

data-dir:    data/
```
This will be rejected by Cabal with `Plain fields are not allowed in between stanzas`.

Thanks to @vlatkoB for reporting this.

PR #281, opened 2018-03-11
Generate valid `.cabal` files when `verbatim` is used top-level (see #280)

sol:
(render top-level verbatim in-between fields and stanzas)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## next
   - Add support for `cxx-options` and `cxx-sources` (see #205)
+  - Generate valid `.cabal` files when `verbatim` is used top-level (see #280)
 
 ## Changes in 0.27.0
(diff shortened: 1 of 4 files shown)


sol:
@vlatkoB I hope this will address it.

vlatkoB:
@sol I patched this to 0.27.0 and tested with this package.yaml
```
# package.yaml
name:        test
version:     '0.1.0.0'
verbatim:
  - data-dir:      data/x1
  - cabal-version: '>= 2.0'

library:
  exposed-modules:
    - Test.hs
```
and the result is correct now
```
test.cabal
name:           test
version:        0.1.0.0
build-type:     Simple
data-dir:       data/x1
cabal-version:  >= 2.0

library
  exposed-modules:
      Test.hs
  other-modules:
      Paths_test
  default-language: Haskell2010
```
This was one very quick resolution. :-)

Issue #292, opened 2018-07-07
unexpected cabal-version higher than 2.2 cannot be specified as a range


dukedave:
If I try to specify `cxx-options: -std=c++11`  I get:

```

- 19:23:
unexpected cabal-version higher than 2.2 cannot be specified as a range. See https://github.com/haskell/cabal/issues/4899
expecting ".", "-", white space, "&&" or "||"

>= 2.2
```

The referenced issue is https://github.com/haskell/cabal/issues/4899, but I don't know enough to know how that relates to this this. I _guess_ it's something sufficiently new in cabal that it isn't supported in hpack, and to that end:

I see this in the the hpack readme:

> Note: Hpack does not require you to specify a cabal-version manually. When generating a .cabal file, Hpack sets the cabal-version automatically based on the features that are used.

Perhaps it's worth documenting what the latest known working `cabal-version` is (and perhaps warning when going beyond it, as a I presume `2.2` is?).

tfausak:
It used to be that the Cabal file had to have something like this:

```
cabal-version: >= 1.10
```

As of Cabal 2.0, the following syntax was allowed:

```
cabal-version: 2.0 -- essentially means ">= 2.0"
```

As of Cabal 2.2, that syntax was required. hpack needs to choose the appropriate syntax for the given version of Cabal.

sol:
The designated workaround for this would be:

```yaml
verbatim:
  cabal-version: 2.2
```

However, Cabal won't accept that neither as `cabal-version` is now required to be at the beginning of the file.

Not even comments are allowed before `cabal-version` anymore.  This is problematic as `hpack` puts a header comment at the beginning of the file.  It's not yet clear to me what's the best way to address this.

sol:
As I understand it, the following steps will be required to address this:

1. We have to put the `cabal-version` at beginning of the generated cabal file, even before the header.  This will make calculating the file hash more convoluted; it's ugly but solvable.
1. When `cabal-version` is `2.1` or higher, we have to skip the `>=` when rendering.

dukedave:
Also if I manually edit the generated `.cabal` to put `cabal-version: 2.2` at the top, I then get:

```
unexpected Unknown SPDX license identifier: 'BSD3' Do you mean BSD-3-Clause?
```

If I also manually set `license:  BSD-3-Clause` then it works.

I'm not sure other licenses have also been changed.
Shall I open a separate issue for this, or keep it all on this one?

sol:
[...]

PR #294, opened 2018-07-10
Support SPDX license identifiers (see #292)

sol:
- Require `cabal-version: 2.2` when SPDX license identifiers are used
- Map cabal-style licenses to SPDX license identifiers when
  `cabal-version` is 2.2 or higher


src/Hpack/License.hs
```diff
@@ -0,0 +1,48 @@
...
+
+parseLicense :: String -> License SPDX.License
+parseLicense license = case spdxLicense of
+  Right l | isUnknown -> MustSPDX l
+  Left  _ | isUnknown -> DontTouch
+  Right l -> CanSPDX l
+  Left  _ -> case cabalLicense of
+    Right l -> CanSPDX (Cabal.licenseToSPDX l)
+    Left _ -> DontTouch
+  where
```
quasicomputational: I think it might be clearer to write these cases like this:

```
  Right l | isRight cabalLicense -> CanSPDX l
  Right l -> MustSPDX l
  Left  _ -> case cabalLicense of
    Right l -> CanSPDX (Cabal.licenseToSPDX l)
    Left _ -> DontTouch
```

Then we can also get rid of the hard-coded `knocnLicenses` list.

quasicomputational:
CI should be fixed by just bumping the stack resolver to something recent.

src/Hpack/License.hs
sol: Yes, exactly.  Right now we don't touch unknown stuff so that Cabal will fail on typos in SPDX license identifiers.  Otherwise e.g. `GPL-2.0-olny` would be converted to `LicenseRef-GPL-2.0-olny`, which I assume we don't want.

sol:
Anybody knows what the AppVeyor build issue is about? Apparently due to `network`?! @kazu-yamamoto

tfausak:
Using the LTS 12.0 resolver instead of a nightly might fix the `network` build problem on AppVeyor.

sol:
@quasicomputational do you know if a different stackage snapshot as suggested by @tfausak will also help?

quasicomputational:
No, I don't think it will. The problem is that the temporary directory generated by AppVeyor looks like `C:\1\blah` - note the backslashes, which cause problems for the `configure` script. Manually overriding the temporary directory works because the specific problem is a backslash followed by a number; there's a fuller fix in Cabal HEAD (swap backslashes for slashes on Windows), but this is a good workaround on AppVeyor today.

tfausak:
Ah, yup. LTS 12 doesn't fix this problem on AppVeyor: https://ci.appveyor.com/project/TaylorFausak/rattletrap/build/446#L336

kazu-yamamoto:
Very strange. `network` 2.6 should be buildable on AppVeyor: https://ci.appveyor.com/project/eborden/network/build/453

We did not change building system for 2.6.

sol:
> there's a fuller fix in Cabal HEAD (swap backslashes for slashes on Windows), but this is a good workaround on AppVeyor today.

@quasicomputational this sounds scary.  I'm surprised that this wasn't included in a bug fix release yet.

src/Hpack/License.hs
sol: I added one more test to make it more clear what's happening here: https://github.com/sol/hpack/pull/294/commits/4d23731ff811f0391545e3707199516a7719e5ec

sol:
Still failing on Windows, not sure what the gist of it is...

sol:
I oppened #295 for the failing AppVeyor build.

sol:
Merged! @quasicomputational thanks for the review :)

Issue #305, opened 2018-07-14
Add support for additional licenses when infering license


sol:
Currently only MIT, BSD2, BSD3 and BSD4 are recognized when inferring `license`.  This covers about 80% of Hackage.

To get closer to 100% we want to extend this with at least:

* ISC
* GPL-2.0
* GPL-3.0
* LGPL-2.1
* LGPL-3.0
* AGPL-3.0
* MPL-2.0
* Apache-2.0

The current approach uses similarity matching using the Levenshtein Distance.  This is fast enough for `MIT` and `BSD*` but gets slow with more lengthy licenses like GPL or Apache.  For this reason this will require a different approach.

We might want to use the [SPDX License List Matching Guidelines](https://spdx.org/spdx-license-list/matching-guidelines) and the corresponding [license templates](https://github.com/spdx/license-list-XML/).

PR #306, opened 2018-07-15
Add support for GPL, LGPL and AGPL when inferring `license` (see #305)

sol:


sol:
Merged but GitHub seems to has issues today... auto close not working... also got some 500s and Travis does not get notified on all pushes neither...

Issue #254, opened 2018-01-18
Add support for build-tool-depends


RyanGlScott:
http://cabal.readthedocs.io/en/latest/developing-packages.html?highlight=build-tool-depends#pkg-field-build-tool-depends

I found myself wanting this [here](https://github.com/haskell-compat/base-orphans/commit/7af12287c025cddab7af6f37313e98af1f3b227b), but because `hpack` doesn't support `build-tool-depends`, I have to add it to the generated `.cabal` file _post facto_.

sol:
@RyanGlScott thank you for opening this issue.  I would like to get the full picture here first, can you give more context on how this fails.  `hspec-discover` is a transitive dependency of `hspec`.  In the past (with cabal sandboxes and `stack`) it just worked to depend on `hspec` and `hspec-discover` would be on the path when the test suite is build.

Does this still work with `cabal new-build`?

RyanGlScott:
You have to declare a `build-tool` dependency on `hspec-discover` in order for it to build successfully with `hspec-discover` (as demonstrated [in this Travis build](https://travis-ci.org/haskell-compat/base-orphans/jobs/330394844#L770), from a commit where I forgot to add it). `new-build` uses a more declarative approach to executable dependencies than old-`build` does, and won't simply install executables as a side effect of installing the corresponding libraries anymore.

sol:
Not excited!  Even so I understand the rational for this from the perspective of a cabal developer, this means that every user of `hspec` has to add

```
build-tool-depends: hspec-discover:hspec-discover == 2.
```

to their cabal file.

Cabal's design decision makes a lot of sense from a technical point of view, BUT it puts additional burden on the user.  A user already stated that she wants to use `hspec` and it does not make things any more convenient for her to state it multiple times.

Besides that:

- it worked in the past, so it's even harder for me to sell this as a "feature" to the average user
- it actively pushes people to requiring `cabal-version:  >= 2.0`

@snoyberg I think you have been working towards component based builds ala `cabal-2` in stack.  How will stack deal with situation like this in the future?

sol:
@RyanGlScott An other thing I'm puzzled about, do you know if `cabal new-*` still solves the dependencies for all components as one set?  Say does

```
build-depends:
  hspec == 2.4.7
build-tool-depends:
  hspec-discover:hspec-discover
```

imply `hspec-discover == 2.4.7`?

sol:
> install executables onto your PATH as a side effect of installing the corresponding libraries

[...]

PR #311, opened 2018-07-17
Transparently add `build-tool-depends` (see #254)

sol:
(when a package depends on `hspec` or `markdown-unlit`)


sol:
@quasicomputational do you have any feedback on this?  I feel urged to release this ASAP to finally put an end to the `build-tool-depends` drama.

quasicomputational:
Looks sensible to me!

sol:
After giving this quite some consideration I decided against this for the following reasons:

- it's too much magic
- it hard codes specific build tools
- and I'm the author of both of these tools which makes the whole story even more dodgy

I think the way forward is https://github.com/hspec/hspec/pull/368 and `defaults` for `hspec`.

PR #313, opened 2018-07-31
Add support for system-build-tools

sol:


test/EndToEndSpec.hs
```diff
@@ -532,6 +532,50 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+          [i|
+          executable:
+            system-build-tools:
+              ghcjs
+          |] `shouldRenderTo` (executable_ "foo" [i|
+          build-tools:
+              ghcjs
+          |]) {packageCabalVersion = ">= 1.22"}
+
+      context "with an uknown system build tool" $ do
```
soenkehahn: `s/uknown/unknown`

sol: Nice catch!  Fixed.

PR #316, opened 2018-08-08
Extend build-tools so that it subsumes Cabal's build-tools and build-tool-depends

sol:
We accept both qualified names in the form of `<package>:<executable>` and unqualified names in the form of `<executable>`.  The details are in he `README`.

This is a breaking change in that system executables are not accepted as `build-tools` anymore.  https://github.com/sol/hpack/pull/313 adds support for `system-build-tools`.

In addition, packages that

- use a custom setup script that extends the list of know build tools
- take advantage of the fact that `Cabal >= 2.0` accepts arbitrary `build-tools`

might break.

To get a better understanding on how this breaking change might affect existing packages I looked at all the packages on Hackage.

The good news:
> None of the packages on Hackage that use Hpack are affected.

So this discussion is mainly about in-house packages that use `build-tools`.

What follows is a list of all `build-tools` that are used by Hackage packages (both Hpack or plain `.cabal`) together with how often they are used, grouped into three categories:

### Packages that would not break if they were using Hpack

hsc2hs: 86
c2hs: 70
alex: 65
happy: 60
cpphs: 10
hspec-discover: 7
hsx2hs: 3
cgen: 1

### Packages that would break due to a system executable if they were using Hpack

ghc: 11
git: 2
llvm-config: 2
gfortran: 1
gcc: 1
couchdb: 1
mcc: 1
nix-store: 1
nix-instantiate: 1
nix-hash: 1
nix-env: 1
nix-build: 1

### Packages that would break due to a Haskell executable if they were using Hpack

All of these would break because the executable name does not match the package name, so Hpack will require a qualified name.

gtk2hsTypeGen: 16
gtk2hsC2hs: 16
gtk2hsHookGenerator: 15
cabal: 1
grgen: 1
cgen-hs: 1

We can consider something like https://github.com/sol/hpack/pull/315 to smoothen the transition.

Opinions and feedback please!

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,48 @@ values are merged with per section values.
 | `ld-options` | · | | |
 | `dependencies` | `build-depends` | | |
 | `pkg-config-dependencies` | `pkgconfig-depends` | | |
-| `build-tools` | · | | |
+| `build-tools` | [`build-tools`](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-build-tools) and/or [`build-tool-depends`](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-build-tool-depends) | | |
 | `when` | | | Accepts a list of conditionals (see [Conditionals](#conditionals)) |
+
+**`build-tools`: A set of Haskell executables that are needed to build this component**
+
+Each element consists of a *name* and an optional *version constraint*.
+
+The name can be specified in two ways:
+
+1. Qualified: `<package>:<executable>`
+1. Unqualified: `<executable>`
+
+A qualified name refers to an executable named `<executable>` from a
+package named `<package>`.
+
+An unqualified name either refers to an executables in the same package, or if
+no such executable exists it is desugared to `<executable>:<executable>`.
+
+`build-tools` can be specified as a list or a mapping.
+
+Examples:
+```yaml
+build-tools:
+  - alex
+  - happy:happy
+  - hspec-discover == 2.*
+```
+```
+build-tools:
+  alex: 3.2.*
+  happy:happy: 1.19.*
+  hspec-discover: 2.*
+```
+
+When generating a `.cabal` file each element of `build-tools` is either added
+to `build-tools` or `build-tool-depends`.
+
+If the name refers to one of `alex`, `c2hs`, `cpphs`, `greencard`, `haddock`,
+`happy`, `hsc2hs` or `hscolour` then the element is added to `build-tools`,
+otherwise it is added to `build-tool-depends`.
+
+This is done to allow compatibility with a wider range of `Cabal` versions.
 
 #### <a name="library-fields"></a>Library fields
 
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -42,6 +42,7 @@ module Hpack.Config (
 , SourceDependency(..)
 , GitRef
 , GitUrl
+, BuildTool(..)
 , GhcOption
 , Verbatim(..)
 , VerbatimValue(..)
@@ -69,6 +70,8 @@ module Hpack.Config (
 , LibrarySection(..)
 , fromLibrarySectionInConditional
 , formatOrList
+
+, toBuildTool
 #endif
 ) where
 
@@ -110,6 +113,7 @@ import           Hpack.Defaults
 import qualified Hpack.Yaml as Yaml
 import           Hpack.Syntax.DependencyVersion
 import           Hpack.Syntax.Dependency
+import           Hpack.Syntax.BuildTools
 import           Hpack.License
 import           Hpack.CabalFile (parseVersion)
 
@@ -277,7 +281,7 @@ data CommonOptions cSources cxxSources jsSources a = CommonOptions {
 , commonOptionsLdOptions :: Maybe (List LdOption)
 , commonOptionsBuildable :: Maybe Bool
 , commonOptionsWhen :: Maybe (List (ConditionalSection cSources cxxSources jsSources a))
-, commonOptionsBuildTools :: Maybe Dependencies
+, commonOptionsBuildTools :: Maybe BuildTools
 , commonOptionsVerbatim :: Maybe (List Verbatim)
 } deriving (Functor, Generic)
 
@@ -339,7 +343,7 @@ instance (Semigroup cSources, Semigroup cxxSources, Semigroup jsSources) => Semi
   , commonOptionsLdOptions = commonOptionsLdOptions a <> commonOptionsLdOptions b
   , commonOptionsBuildable = commonOptionsBuildable b <|> commonOptionsBuildable a
   , commonOptionsWhen = commonOptionsWhen a <> commonOptionsWhen b
-  , commonOptionsBuildTools = commonOptionsBuildTools b <> commonOptionsBuildTools a
+  , commonOptionsBuildTools = commonOptionsBuildTools a <> commonOptionsBuildTools b
   , commonOptionsVerbatim = commonOptionsVerbatim a <> commonOptionsVerbatim b
   }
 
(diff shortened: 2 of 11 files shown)


src/Hpack/Config.hs
```diff
@@ -1265,9 +1275,11 @@ toSection = go
       , sectionDependencies = fromMaybe mempty commonOptionsDependencies
       , sectionPkgConfigDependencies = fromMaybeList commonOptionsPkgConfigDependencies
       , sectionConditionals = map toConditional (fromMaybeList commonOptionsWhen)
-      , sectionBuildTools = fromMaybe mempty commonOptionsBuildTools
+      , sectionBuildTools = maybe mempty toBuildToolMap commonOptionsBuildTools
```
quasicomputational: `maybe mempty f` === `foldMap f`, which may or may not be an improvement

quasicomputational: Also, this silently has first-one-wins behaviour when the same build tool is specified multiple times. Worth documenting if nothing else.

test/EndToEndSpec.hs
```diff
@@ -469,6 +469,69 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+          [i|
+          build-tools:
+            - bar == 0.1.0
+          executables:
+            bar:
+              build-tools:
+                - foo:bar == 0.2.0
+          |] `shouldRenderTo` executable_ "bar" [i|
+          build-tools:
+              bar ==0.2.0
```
quasicomputational: This test doesn't seem to match up with its name.

src/Hpack/Config.hs
```diff
@@ -1276,6 +1288,15 @@ toSection = go
       where
         conditional (Condition (Cond c)) = Conditional c
 
+toBuildTool :: String -> [String] -> ParseBuildTool -> BuildTool
+toBuildTool packageName_ executableNames = \ case
+  QualifiedBuildTool pkg executable
+    | pkg == packageName_ && executable `elem` executableNames -> LocalBuildTool executable
```
quasicomputational: This might go wrong if people define an executable in a verbatim section, but I guess that using verbatim means that you know what you're doing. Possibly worth calling out this case in the docs, as a hint for why hpack would be doing the 'wrong' thing.

test/EndToEndSpec.hs
```diff
@@ -469,6 +469,69 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+        [i|
+        executable:
+          build-tools:
+            hspec-discover: 0.1.0
+        |] `shouldRenderTo` (executable_ "foo" [i|
+        build-tool-depends:
+            hspec-discover:hspec-discover ==0.1.0
+        |]) {
+          -- NOTE: We do not set this to 2.0 on purpose, so that the .cabal
+          -- file is compatible with a wider range of Cabal versions!
```
quasicomputational: What's the meaning of 'compatible' here? Will the `.cabal` file actually work as desired with old Cabals, or will the file be read without complaints but the tool not be run? If it's the latter, I think we should put a proper constraint on.

README.md
```diff
@@ -179,9 +179,49 @@ values are merged with per section values.
...
+```yaml
+build-tools:
+  - alex
+  - happy:happy
+  - hspec-discover == 2.*
+```
+```
+build-tools:
+  alex: 3.2.*
+  happy:happy: 1.19.*
```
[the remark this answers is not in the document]
sol: I think hat is understandable, yes. In this specific case a user can just use:

```yaml
build-tools:
  - happy: 1.19.*
```

In cases where the colon is required I would encourage users to use list syntax instead:

```yaml
build-tools:
  - happy:happy == 0.19.*
```

So the remaining question is whether we should make something like `happy:happy` explicitly invalid.  I'm leaning towards "no", but I'm interested in other opinions.

src/Hpack/Config.hs
sol: It should be last one wins?!

test/EndToEndSpec.hs
sol: Huh? For what reason?

src/Hpack/Config.hs
sol: Regarding docs, how about https://github.com/sol/hpack/pull/323?

test/EndToEndSpec.hs
sol: As per https://cabal.readthedocs.io/en/latest/developing-packages.html#buildtoolsbc, for older versions of `cabal-install` a user has to make sure that required build tools are on the `PATH`.

The specific case of `hspec-discover` will work with older versions of `cabal-install` without any additional user effort, as `hspec-discover` is installed as a dependency of `hspec`.

src/Hpack/Config.hs
quasicomputational: The `Monoid` instance of `Map` uses `union`, which is left-biased. If it's going to be last-one-wins, then we'll have to use `Dual` or `reverse` the input, then.

test/EndToEndSpec.hs
quasicomputational: Oh, sorry, I got confused. I thought that it was referring to a non-local executable but I see that it's doing the right thing.

test/EndToEndSpec.hs
quasicomputational: OK; I think that that definitely needs a mention in the docs. `cabal check` should also be complaining about this because `build-tool-depends` is new in 2.0, but I see that it actually isn't, but we can't rely on that.

src/Hpack/Config.hs
sol: `BuildTools` are still a list when the concatenation happens.   So the `Monoid` instance for `[]` is used.  This is to make sure that the cases

```
gives per-section unqualified names precedence over global qualified names
```
and
```
gives per-section qualified names precedence over global unqualified names
```
are met.

These two tests also test that we indeed have last one wins behavior.

test/EndToEndSpec.hs
sol: > OK; I think that that definitely needs a mention in the docs. 

What exactly do you think should we add to the docs?

> `cabal check` should also be complaining about this because `build-tool-depends` is new in 2.0, but I see that it actually isn't, but we can't rely on that.

I think this is by design for the purpose of backwards compatibility.  This is at least how I read https://cabal.readthedocs.io/en/latest/developing-packages.html#buildtoolsbc.

quasicomputational: https://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-build-tool-depends says that it's ignored with a warning on old versions of Cabal.

Suggested docs warning:

> Note that packages using `build-tool-depends` will likely not work correctly with versions of Cabal prior to 2.0, because the build tools will not be run.

This is why I was asking about what counted as 'compatibility': packages that wind up using `build-tool-depends` will likely break when they're built on old Cabals, unless they're using the build tools for something non-essential; in a real sense they're not compatible with old Cabals, even if those old Cabals can read the rest of the package description. ISTM we'd be better off giving those `cabal-version: 2.0` and dodging both future warnings and breakage.

sol: > packages that wind up using `build-tool-depends` will likely break when they're built on old Cabals

The large majority of packages that use `build-tool-depends` use `hspec-discover`.  At least these packages will continue to work with old versions of `cabal-install`.  The same is true for packages that use `markdown-unlit`.

For other packages that use `build-tool-depends`, they may work as well:

1. either because the executable is pulled in as a dependency (as is the case for `hspec-discover`)
1. or because the executable is on the `PATH` by other means

So for many packages there is a good chance that a package that uses `build-tool-depends` will work with an old version of `cabal`.

If it does not work, `cabal` will fail due to the missing executable at build time.  If we would require `cabal-version: 2.0` it would fail at configuration time.

Failing at configuration time is strictly better than failing at build time. But the question is still which of the following options do we prefer:

1. Have the majority of packages that use `build-tool-depends` work with old versions of `cabal` and the ones that don't work fail with a build error
1. Have all packages that use `build-tool-depends` fail at configuration time with old versions of `cabal`

> ISTM we'd be better off giving those `cabal-version: 2.0` and dodging both future warnings and breakage.

I don't expect there to be any future warnings.  As I understand it `build-tool-depends` was carefully implemented to allow for this kind of backwards compatibility, arguably at the cost of comprehensibility.

PR #321, opened 2018-08-15
Allow dependency modules to be hidden, exposed and renamed.

### LINKED ISSUE #318: Proposal: module hiding and renaming parts of mixins
The [`mixins`](https://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-mixins) field is useful for renaming and hiding (or selectively exposing) modules from dependencies, even without supporting the full power of backpack. To use it today with hpack, you need to use `verbatim`, and it's awkward in a couple of ways:

* You have to repeat the package in the `depends` field as well as the verbatim `mixins` field. Those two are not necessarily close together, increasing the chance of bugs.
* Some of Cabal's syntax isn't great; e.g., to get renaming plus default passthrough, you have to write two lines: `foo (Bar as Baz), foo hiding (Bar)`.
* You likely have to remember to set the right `cabal-version` with `verbatim`.

I think hpack could usefully support the module renaming/hiding parts and punt on supporting `requires`; with `verbatim`, anyone who needs that power has an escape hatch to get at it.

My proposed syntax is this: we extend object dependencies to take an object instead of just a string version, leading to something looking like this:

```
  depends:
    base:
      version: ">= 4.10.1.0 && < 4.12"
      renaming:
        - from: Some.Long.Name
          to: Short
      hiding:
        - A
    blah:
      exposing:
        - Blah
```

This will generate something like this:

```
  build-depends:
    base >= 4.10.1.0 && < 4.12,
    blah
  mixins:
    base (Some.Long.Name as Short),
    base hiding (A),
    blah (Blah)
```

It would be an error to have both `hiding` and `exposing` keys. The default should be to expose all modules of the dependency, even if some modules are renamed; `exposing: []` can be written if the opposite behaviour is wanted.

I'm up for implementing this, but I'd like to hear if anyone thinks this design or the UX could be improved.

quasicomputational:
When used, this will produce `mixins` fields in the generated package
description.

Along the way, defining individual dependencies as objects has become
more flexible: they can now be a objects with `name` and `version`
keys. Some of this was present before due to `SourceDependency`, but
now it is more flexible and is useful generally, rather than only for
integration with tinc.

Closes #318.


src/Hpack/Syntax/Dependency.hs
```diff
@@ -34,33 +39,92 @@ instance FromValue Dependencies where
...
+    Null -> return $ DependencyInfo AnyVersion Nothing defaultModuleExposure []
+    Number n -> do
+      let version = scientificToDependencyVersion n
+      return $ DependencyInfo version Nothing defaultModuleExposure []
+    String s -> do
+      let input = T.unpack s
+      version <- parseVersionRange ("== " ++ input) <|> parseVersionRange input
+      return $ DependencyInfo version Nothing defaultModuleExposure []
+    Object o -> dependencyInfoFromObject o
+    _ -> typeMismatch "Null, Object, Number, or String" v
```
quasicomputational: Ideally I'd like to share more code with the `FromValue DependencyVersion` instance, but I can't quite get it to work without regressing the error message here.

src/Hpack/Syntax/DependencyVersion.hs
```diff
@@ -5,10 +5,13 @@ module Hpack.Syntax.DependencyVersion (
...
 , GitUrl
 , DependencyVersion(..)
-, SourceDependency(..)
 , dependencyVersionFromCabal
+, SourceDependency(..)
+, sourceDependencyFromObject
 
 , scientificToVersion
+, scientificToDependencyVersion
+, parseVersionRange
```
quasicomputational: Might think about re-organising `H.S.Dependency` and `H.S.DependencyVersion` now; there's not a whole lot of logic about what's going in which module left.

README.md
```diff
@@ -232,6 +232,65 @@ values are merged with per section values.
 | `manual` | · | | Required (unlike Cabal) |
 | `default` | · | | Required (unlike Cabal) |
 
+#### <a name="dependencies"></a> Dependencies
+
+Dependencies can be specified as either a list or an object. These are
+equivalent:
+
+```
+  depends:
```
sol: Again, this is not a proper review yet, but just that we don't forget latter on, this should be `dependencies` instead of `depends`.

Other than that, the docs look nice :).

quasicomputational:
OK, I've implemented the new plan from #318. I've tentatively got this down as a breaking change because the new types leak out through `Hpack.Config`, so the README says 'since 0.31'.

src/Hpack/Render/Dependency.hs
```diff
@@ -0,0 +1,49 @@
...
+import           Data.Foldable (toList)
+import qualified Data.Map as Map
+import           Data.Semigroup (Semigroup(..))
+import           Hpack.Render.Dsl
+import           Hpack.Syntax.Dependency
+import           Hpack.Syntax.DependencyVersion
+
+data DependencyFields = DependencyFields {
+  _dependencyFieldDepends :: DList String
+, _dependencyFieldMixins :: DList String
```
sol: Does the `DList` actually makes a difference here?  Does `foldMapWithKey` give us left associative concatenation?

src/Hpack/Render/Dependency.hs
```diff
@@ -0,0 +1,49 @@
...
+     , [ Field "mixins" $ CommaSeparatedList $ toList mixins
+       | not (null (toList mixins))
+       ]
+     ]
+
+renderDependency :: String -> DependencyInfo -> DependencyFields
+renderDependency name (DependencyInfo version mixins) =
+  DependencyFields
+    (DList.singleton $ name ++ renderVersion version)
+    (DList.fromList [ name ++ " " ++ mixin | mixin <- mixins ])
```
sol: This code is sophisticated and clever.

But if we remove the indirection over the type class it may be easier to understand for the casual reader.  It's also much shorter:
```haskell
renderDependencies :: String -> Dependencies -> [Element]
renderDependencies name deps = [
    Field name (CommaSeparatedList renderedDeps)
  , Field "mixins" (CommaSeparatedList $ concat mixins)
  ]
  where
    (renderedDeps, mixins) = unzip . map renderDependency . Map.toList $ unDependencies deps

renderDependency :: (String, DependencyInfo) -> (String, [String])
renderDependency (name, DependencyInfo version mixins) = (
      name ++ renderVersion version
    , [ name ++ " " ++ mixin | mixin <- mixins ]
    )
```

src/Hpack/Render/Dependency.hs
```diff
@@ -0,0 +1,49 @@
...
+  mappend = (<>)
+
+renderDependencies :: String -> Dependencies -> [Element]
+renderDependencies name deps =
+  let DependencyFields renderedDeps mixins =
+        Map.foldMapWithKey renderDependency $ unDependencies deps
+  in concat
+     [ [ Field name $ CommaSeparatedList $ toList renderedDeps ]
+     , [ Field "mixins" $ CommaSeparatedList $ toList mixins
+       | not (null (toList mixins))
```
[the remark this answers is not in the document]
sol: Ah, I see it's necessary for `Hpack.Render.DependencySpec` to pass.

What I said is still true for the overall program, as empty fields are omitted when rendering.

test/Hpack/Syntax/DependencySpec.hs
```diff
@@ -73,36 +76,20 @@ spec = do
...
               path: ../hpack
-          |] `shouldDecodeTo_` Dependencies [("hpack", SourceDependency source)]
-
-        context "when ref is missing" $ do
-          it "produces accurate error messages" $ do
-            [yaml|
-              - name: hpack
-                git: sol/hpack
-                ef: master
-            |] `shouldDecodeTo` left "Error while parsing $[0] - key \"ref\" not present"
```
sol: I'm eager to keep this error messages as accurate as possible.

sol:
`version` is also useful for source dependencies.  That way you can include a version constraint in the generated cabal file, even if you are using source dependencies.  For that reason I made `DependencyVersion` now a product of `SourceDependency` and `VersionConstraint`.

quasicomputational:
Funny story, my PR originally did that (turning it into a product), but then I decided to back that out to minimise the diff.

I'll get cracking on the rebase now and see how it goes.

test/Hpack/Syntax/DependencySpec.hs
quasicomputational: The current PR version (pre-rebase) is going to emit an 'unused field' warning for these. (Or it should. I'll re-add tests.) Is that good enough, or does it have to be a hard error?

src/Hpack/Render.hs
```diff
@@ -239,10 +240,10 @@ renderSection renderSectionData extraFieldsStart extraFieldsEnd Section{..} = ad
   , renderDirectories "extra-frameworks-dirs" sectionExtraFrameworksDirs
   , Field "frameworks" (LineSeparatedList sectionFrameworks)
   , renderLdOptions sectionLdOptions
-  , renderDependencies "build-depends" sectionDependencies
   , Field "pkgconfig-depends" (CommaSeparatedList sectionPkgConfigDependencies)
   ]
   ++ renderBuildTools sectionBuildTools sectionSystemBuildTools
+  ++ renderDependencies "build-depends" sectionDependencies
```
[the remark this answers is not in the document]
quasicomputational: Sorry, don't understand this comment. We shouldn't accept objects with those keys as `build-tools`?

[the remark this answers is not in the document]
sol: Yes, current `master` already does that.

test/Hpack/Syntax/DependencySpec.hs
sol: It's a hard error with current `master`.  We just keep the exiting tests in place.

src/Hpack/Render.hs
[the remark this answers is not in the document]
quasicomputational: OK. So nothing to do here? I'm not going to need to touch the specific `build-tools` bits at all and only a little bit of refactoring in `H.S.DependencyVersion` is the current plan.

quasicomputational:
That wasn't too terrible, especially for a rebase with a file move in it.

Thanks for the review so far - I agree that the code was awkwardly duplicated across the various places that wanted to parse versions (which this PR was only making worse) and this has definitely cleaned it up. Now all that logic's in exactly one place and hopefully it'll stay there!

test/Hpack/Render/DependencySpec.hs
```diff
@@ -0,0 +1,52 @@
...
+        [ Field "build-depends" $ CommaSeparatedList
+           [ "bar"
+           , "foo"
+           ]
+        , Field "mixins" $ CommaSeparatedList
+            [ "bar hiding (Spam)"
+            , "bar (Spam as Spam1) requires (Mod as Sig)"
+            , "foo (Foo as Foo1)"
+            ]
+        ]
```
sol: We don't have `Hpack.Render.Dependency` anymore, right? If so, then this should be merged into `Hpack.RenderSpec`.

sol: To motivated this: I have editor macros to jump from source to tests, for that reason I try to follow a strict convention on where to find tests for a specific piece of code.

sol:
@quasicomputational thanks a lot!

quasicomputational:
You're welcome!

Are there other plans for 0.31? It'd be pretty nifty if this could roll out with the next stack version, which is just around the corner, but I think that a hpack version bump wouldn't be a problem for them.

PR #325, opened 2018-08-29
Keep declaration order for literal files in c-sources (fixes #324)

### LINKED ISSUE #324: Preserve ordering in lists
For whatever reason, `cabal new-repl` requires `c-files` to be in dependency order. However, currently hpack breaks the ordering (cabal file has all entries in alphabetic order). Am I missing some option to preserve custom ordering for `c-files` field?

sol:


test/EndToEndSpec.hs
```diff
@@ -681,6 +681,19 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
         |]
 
     describe "c-sources" $ before_ (touch "cbits/foo.c" >> touch "cbits/bar.c") $ do
+      it "keeps declaration order" $ do
+        -- IMPORTANT: This is crucial as a workaround for https://ghc.haskell.org/trac/ghc/ticket/13786
+        [i|
+        c-sources:
+          - cbits/foo.c
+          - cbits/bar.c
```
quasicomputational: Might suggest an additional entry here, to make sure it really is respecting ordering and not (e.g.) outputting the input in reverse-sorted order.

quasicomputational: Also I guess the intersection of order preservation and duplicate resolution will need a test. I am not sure what the right behaviour is here.

sol: > Might suggest an additional entry here, to make sure it really is respecting ordering and not (e.g.) outputting the input in reverse-sorted order.

I don't think it's strictly necessary in terms of TDD, but I adapted the tests ;)

> Also I guess the intersection of order preservation and duplicate resolution will need a test. I am not sure what the right behaviour is here.

The correct behavior would probably be to give explicitly mentioned stuff precedence over stuff from expanded glob patterns.  Implementing this, while possible, will make the code quite a bit more complicated.  Two observations:

1. As things are it's probably not a good idea to use globs with `c-sources`
2. Right now things that you list first take precedence over things that you list later

I think (1) ad (2) will allow users to work around the `c-sources` stuff.  But other than that, I'll leave this unspecified.

PR #326, opened 2018-09-02
Exit successfully in more cases.

quasicomputational:
Previously, `hpack` would return with a failure exit code if it
detected that the `.cabal` file had been previously generated by a
newer version of `hpack` - even if the `.cabal` file was precisely
what would have been generated anyway!

This is an usability improvement usability whenever there are multiple
versions of `hpack` in use in a single environment (e.g., one version
baked into `stack`, and a stand-alone version being used for
experimentation and evaluation of new features).


src/Hpack.hs
```diff
@@ -132,6 +132,7 @@ data Status =
     Generated
   | ExistingCabalFileWasModifiedManually
   | AlreadyGeneratedByNewerHpack
+  | GeneratedByNewerHpackButUnchanged
```
sol: I'm tempted to just treat this as `OutputUnchanged` and not add a new constructor here.

If we decide to add the new constructor, we will also need corresponding PRs for `stack`, `cabal2nix` and `tinc`.

quasicomputational: Makes sense. Let's go with no new constructor, since I'm not sure any API consumers will care about the difference.

sol:
Thanks a lot!

PR #332, opened 2018-09-13
Show the header when printing to stdout as well

### LINKED ISSUE #331: hpack doesn't create hash header when told to print to stdout
Use case:

* docker with project mounted read-only into the container
* shake has a rule to create the project .cabal file by running the container and writing the stdout result to the .cabal file on the host (because stack is automagic and doesn't let you control anything, e.g. how hpack is invoked)

Except, this doesn't work, because the hpack header is missing, resulting in the host stack to complain about a modified .cabal file all the time.

Why is the header missing? Is there a reason?

hasufell:
Fixes #331

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -181,9 +181,10 @@ hpackResultWithVersion v (Options options force toStdout) = do
   case status of
     Generated -> do
       let hash = sha256 withoutHeader
+          out  = cabalVersion ++ header (decodeOptionsTarget options) v hash ++ body
       if toStdout
-        then Utf8.putStr withoutHeader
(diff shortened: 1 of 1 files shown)


src/Hpack.hs
```diff
@@ -181,9 +181,10 @@ hpackResultWithVersion v (Options options force toStdout) = do
   case status of
     Generated -> do
       let hash = sha256 withoutHeader
+          out  = cabalVersion ++ header (decodeOptionsTarget options) v hash ++ body
       if toStdout
```
ip1981: No reason to keep this.

sol:
Thanks a lot!

sol:
Released to Hackage as `0.31.1`.

PR #335, opened 2018-10-21
Make GitHub recognize the MIT License

### LINKED ISSUE #334: License not recognized as BSD-3-Clause.
The [license](https://github.com/sol/hpack/blob/master/LICENSE) is not being recognised by github as BSD-3-Clause. For an example that is recognised, see the [license of hpack-dhall](https://github.com/BlockScope/hpack-dhall/blob/master/LICENSE).

tfausak:
Fixes #334, which mentions BSD-3-Clause even though this repository uses MIT.

This uses the text from <https://choosealicense.com/licenses/mit/>.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+MIT License
+
 Copyright (c) 2014-2018 Simon Hengel <sol@typeful.net>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


sol:
@tfausak thanks for the PR.  I'm actually not sure why GitHub does not recognize the license, but I think it's not due to the actual license text.  I use the exact same license text for most of my projects, and it is recognized as MIT in other places, e.g. https://github.com/sol/doctest/blob/master/LICENSE.

As such I'm tempted to conclude that this is a GitHub bug rather than something want to address by change the license text.

sol:
Resolved by GitHub support, closing.

PR #336, opened 2018-10-21
Upgrade to Stackage LTS 12.0

### LINKED ISSUE #329: Fix AppVeyor tests

tfausak:
This fixes #329. Or it should, at least.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,7 @@ clone_folder: "c:\\stack"
 environment:
   global:
     STACK_ROOT: "c:\\sr"
+    TMP: "c:\\tmp"
 
 test_script:
 - stack setup > nul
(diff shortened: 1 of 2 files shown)


sol:
If we can get the windows build working again that would be awesome.  But still seems to be failing, not sure what the issue is.

tfausak:
I'm not entirely sure why it failed either. I definitely expected it to succeed, but CI makes fools of us all.

quasicomputational:
AIUI, we'd need to use either GHC 8.6.n or 8.4.4 to fix AppVeyor. So this is blocked on 8.6.2 being released (since 8.6.1 has a horrible bug) and then Stackage upgrading. (Or we could build our own snapshot as a hack, I guess.)

sol:
@quasicomputational so `lts-12.16` which uses GHC 8.4.4 should work then?

@tfausak can you try if upgrading to `lts-12.16` help?

tfausak:
Somehow it's even less happy with LTS 12.16 than it was with LTS 12.0. 🤔

sol:
Fixed on `master`.

PR #345, opened 2018-12-13
Make CI green

quasicomputational:


quasicomputational:
Well, at least it's getting further than before, and now we have an exciting new error to puzzle out...

quasicomputational:
Looks like we're blocked on a new `clock` release. corsis/clock#52 is the issue and it's been fixed in their HEAD, but that's not made it to Hackage yet (corsis/clock#55).

quasicomputational:
One thing gets fixed and another breaks, within the same day. With any luck, this will green up CI across the board.

quasicomputational:
Ping @sol.

sol:
@quasicomputational thanks a lot!  You are great!

PR #346, opened 2018-12-19
Add `--check-only`, for staleness checking.

quasicomputational:
With this option, hpack will not generate the `.cabal` file, but will
return with a non-zero status if the `.cabal` file is missing, stale,
or manually modified.


quasicomputational:
Ping?

sol:
@quasicomputational sorry for never getting back at this.  This was at a time when I was struggling with RSI badly.  That's the reason.  However, I think we want something like this.  Do you want to pick it up again?  If not, I'll look into it.

sol:
I haven't looked at the code, but from a users perspective what I would want to have is something that gives me a non-negative exit code if `.cabal` is stale.  Maybe have `--check` or `--dry-run` print a status, but when combined with `--silent` (should that be `--quiet` actually?), have it indicate the status by its exit code.

PR #348, opened 2018-12-27
Update README.md

flip111:
add info about extensions

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,8 +159,8 @@ values are merged with per section values.
 | --- | --- | --- | --- |
 | `buildable` | · | | Per section takes precedence over top-level |
 | `source-dirs` | `hs-source-dirs` | | |
-| `default-extensions` | · | | |
-| `other-extensions` | · | | |
+| `default-extensions` | · | | extensions that will be enabled in every source file |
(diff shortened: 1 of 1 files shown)


andreasabel:
Should this simple PR be merged or closed?

sol:
I'm not sure.  I think it is a good thing if the Hpack docs are useful on their own. On the other hand, for stuff that is almost literally mapped to `.cabal` it might make sense to reference the Cabal docs instead.  Thoughts?

andreasabel:
> I'm not sure. I think it is a good thing if the Hpack docs are useful on their own. On the other hand, for stuff that is almost literally mapped to `.cabal` it might make sense to reference the Cabal docs instead. Thoughts?

Certainly.
But the doc in this PR is better than no doc, so it can't hurt.

mpilgrem:
Problems with a large number of links to the Cabal package description specification are (a) linking to which version of the specification? And (b) the 'cost' of maintaining the links.

Rather that treat these two Hpack keys differently from all the others, I think a better solution is:
* provide one introductory link in this document to the Cabal User Guide, and point the reader to the relevant part; and
* state expressly that a `.` in the `Cabal` column means that the Hpack key corresponds directly to a field in the Cabal package description of the same name.

Issue #342, opened 2018-11-07
Comma in filenames


jawsdark:
Cabal needs path/filenames with commas to be enclosed in double quotes.
```
data-files:
  "path/some,file"
```
If the quotes are missing, then Cabal reads it as two different files:
```
data-files:
  path/some,
  file
```
If I use glob patterns like `path/*` the generated file contains `path/some,file` without quotes and the problem occurs. I did not find any way to enforce double quotes on path/filenames in the generated Cabal file. This could be solved by just putting double quotes around all files listed in the generated Cabal file.

sol:
Hey! Thanks for reporting this.  As a workaround you probably need to specify those files by hand for now.

To fix this proper I think we should implement Cabal's escaping rules properly.  The `Cabal` library may already provide something that we can use.  PR welcome!

googleson78:
This is now fixed in 89ecd99.

Issue #350, opened 2019-01-21
Spaces in filenames using wildcards in extra-source-files doesn't work properly


googleson78:
When I include a glob pattern in `extra-source-files`, if the glob matches a file with spaces in it, hpack doesn't escape the spaces in the generated cabal file, causing cabal to not find it.

googleson78:
Fixed in 89ecd99

PR #351, opened 2019-02-01
Quote filenames generated by a glob, which have special characters (whitespace, commas) in them

googleson78:
This is in relation to #350.

I also tried to escape the spaces or to use single quotes, but cabal doesn't like either of those.

### DIFF
--- a/src/Hpack/Util.hs
+++ b/src/Hpack/Util.hs
@@ -132,12 +132,7 @@ expandGlobs name dir patterns = do
     quoteSpecial :: FilePath -> FilePath
     quoteSpecial fp
       | any (\x -> isSpace x || x == ',') fp
-        = let escapeSpecial :: Char -> String
-              escapeSpecial '\n' = "\\n"
-              escapeSpecial '"'  = "\\\""
-              escapeSpecial '\\'  = "\\\\"
-              escapeSpecial x    = [x]
-          in "\"" ++ concatMap escapeSpecial fp ++ "\""
+        = show fp
       | otherwise      = fp
 
     normalize :: FilePath -> FilePath
--- a/src/Hpack/Util.hs
+++ b/src/Hpack/Util.hs
@@ -130,9 +130,12 @@ expandGlobs name dir patterns = do
           | otherwise = []
 
     quoteSpaces :: FilePath -> FilePath
-    quoteSpaces fp = if any isSpace fp
-                     then "\"" ++ fp ++ "\""
-                     else fp
+    quoteSpaces fp
+      | any isSpace fp = let escapeNewline :: Char -> String
+                             escapeNewline '\n' = "\\n"
+                             escapeNewline x    = [x]
+                         in "\"" ++ concatMap escapeNewline fp ++ "\""
+      | otherwise      = fp
 
     normalize :: FilePath -> FilePath
     normalize = toPosixFilePath . makeRelative dir
--- a/src/Hpack/Util.hs
+++ b/src/Hpack/Util.hs
@@ -131,10 +131,12 @@ expandGlobs name dir patterns = do
 
     quoteSpaces :: FilePath -> FilePath
     quoteSpaces fp
-      | any isSpace fp = let escapeNewline :: Char -> String
-                             escapeNewline '\n' = "\\n"
-                             escapeNewline x    = [x]
-                         in "\"" ++ concatMap escapeNewline fp ++ "\""
+      | any isSpace fp
+        = let escapeSpecial :: Char -> String
+              escapeSpecial '\n' = "\\n"
+              escapeSpecial '"'  = "\\\""
+              escapeSpecial x    = [x]
+          in "\"" ++ concatMap escapeSpecial fp ++ "\""
       | otherwise      = fp
 
     normalize :: FilePath -> FilePath
--- a/src/Hpack/Util.hs
+++ b/src/Hpack/Util.hs
@@ -135,6 +135,7 @@ expandGlobs name dir patterns = do
         = let escapeSpecial :: Char -> String
               escapeSpecial '\n' = "\\n"
               escapeSpecial '"'  = "\\\""
+              escapeSpecial '\\'  = "\\\\"
               escapeSpecial x    = [x]
           in "\"" ++ concatMap escapeSpecial fp ++ "\""
       | otherwise      = fp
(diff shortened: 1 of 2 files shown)


quasicomputational:
For reference, [the Cabal docs](https://cabal.readthedocs.io/en/latest/developing-packages.html#package-descriptions) describe the syntax.

Two cases that this patch doesn't handle but probably ought to:
* Newline in filename
* Double quote in filename

Cases I'd like to see tested:
* Single quote in filename
* Leading non-space whitespace (e.g., tabs)

This isn't going to be a panacea because Cabal has no way to escape globs, but it's an improvement, and I guess spaces in filenames are common enough that they shouldn't bite people.

quasicomputational:
Ah, backslashes will also need escaping, since they're legal POSIX path characters and Windows's path component separators.

googleson78:
Currently given the file `'hehe\ kakakaka'`, `hpack` outputs `../scenes/hehe\\ kakakaka`, which when surrounded with quotes I would expect to work. *However* it doesn't. I do not understand this.
So
```
"../scenes/hehe\\ kakakaka"
```
outputs
```
Warning: File listed in benchmark.cabal file does not exist: ../scenes/hehe\ kakakaka
```

However
```
"../scenes/hehe\\\\ kakakaka"
```
is fine, there is no warning.

I created the file with
```
touch 'hehe\\ kakakaka'
```
(in `zsh`, don't know if it matters).

I assume that either
* I am stupid
* this is some `cabal` peculiarity

Obviously I can make this work by also escaping `\`, but I am asking if you could link something relevant so that I don't do something stupid.

googleson78:
Currently given the file `'hehe\ kakakaka'`, `hpack` outputs `../targetdir/hehe\\ kakakaka`, which when surrounded with quotes I would expect to work. *However* it doesn't. I do not understand this.
So
```
"../targetdir/hehe\\ kakakaka"
```
outputs
```
Warning: File listed in blabla.cabal file does not exist: ../targetdir/hehe\ kakakaka
```

However
```
"../targetdir/hehe\\\\ kakakaka"
```
is fine, there is no warning.

I created the file with
```
touch 'hehe\\ kakakaka'
```
(in `zsh`, don't know if it matters).

I assume that either
* I am stupid
* this is some `cabal` peculiarity

Obviously I can make this work by also escaping `\`, but I am asking if you could link something relevant so that I don't do something stupid.

(in my `package.yaml` I have `../targetdir/*` listed in `extra-source-files`)

quasicomputational:
What command exactly is producing that error?

googleson78:
```
stack build --fast --file-watch
```

results in
```
Warning: File listed in blabla.cabal file does not exist: ../targetdir/hehe\\

Warning: File listed in blabla.cabal file does not exist: kakakaka
```

after which I modify the `blabla.cabal`
From:
```
../targetdir/hehe\\ kakakak
```
to
```
"../targetdir/hehe\\ kakakak"
```
after which
```
stack build --fast --file-watch
```
Which results in:
```
Warning: File listed in blabla.cabal file does not exist: ../targetdir/hehe\ kakakaka
```

quasicomputational:
Pushed it to https://github.com/quasicomputational/extra-source-files-backslash

`cabal new-build all` works fine without warning and `new-sdist` generates a fine-looking tarball, so I suspect a Stack bug.

googleson78:
Yeah, it's not the box, it's my brain malfunctioning.

I did
```
touch 'hehe\\ kakakaka'
```
which, obviously creates a file with **two** backslashes in it...

Sorry for the waste of time :(

In any case I also fixed backslash escaping. I'm going to push it now.

googleson78:
Should I also fix #342 here?

quasicomputational:
Good idea. Should be easy now, yeah.

src/Hpack/Util.hs
```diff
@@ -122,13 +122,24 @@ expandGlobs name dir patterns = do
...
 
+    quoteSpecial :: FilePath -> FilePath
+    quoteSpecial fp
+      | any (\x -> isSpace x || x == ',') fp
+        = let escapeSpecial :: Char -> String
+              escapeSpecial '\n' = "\\n"
+              escapeSpecial '"'  = "\\\""
+              escapeSpecial '\\'  = "\\\\"
+              escapeSpecial x    = [x]
+          in "\"" ++ concatMap escapeSpecial fp ++ "\""
```
sol: As I understand it, we can just use `show` here.

src/Hpack/Util.hs
```diff
@@ -122,13 +122,24 @@ expandGlobs name dir patterns = do
     fromResult :: GlobResult -> ([String], [FilePath])
     fromResult (GlobResult pattern compiledPattern files) = case files of
       [] -> (warning, literalFile)
-      xs -> ([], map normalize xs)
+      xs -> ([], map (quoteSpecial . normalize) xs)
```
sol: I think quoting is a representational concern.  As such, I would tend to do this at the point we actually render the `.cabal` file.  Does this make sense?

What we do want to do here though is `newtype`-wrap `FilePath` so that the type checker helps us to ensure that we actually do the quoting at all relevant places.

sol:
@googleson78 @quasicomputational sorry for joining the party late.  Big thanks to the two of you for taking care of this.  I added some notes that I think we should address before this is good to go.

src/Hpack/Util.hs
googleson78: Should I add a "quotable" string constructor to `Hpack.Render.Dsl.Value`? Or should I quote them before then?

If I take the first approach (which I like more), then the issue becomes:
`LineSeperatedList` must also be able to contain "quotable" strings. Which will cause a lot of unneeded wrapping/unwrapping.

But if I quote them in `Hpack.Render` and pass them in like they are now, that would defeat the whole point, at least the way I understand it.

I chose to try out the second approach. You can see how it looks in this pr: https://github.com/googleson78/hpack/pull/1/files

It seems to be working(TM), the tests don't compile, but I just wanted to see what you think first.

googleson78:
Ping (?) because I'm not sure if you were actually notified by the last changes here.

sol:
I ended up with https://github.com/sol/hpack/commit/89ecd9962ec2d258160119cc185afd5f1c6b5a2f on top of your existing changes. Released to Hackage, thanks a lot for bringing this up.

PR #363, opened 2019-07-19
Update package.yaml

amigalemming:
add README.md such that it shows up on Hackage


sol:
Hey, thanks for your contribution!

One concern with this change is that this will render the `README` on Hackage.  I don't want to put in the effort to make sure the `README` renders on Hackage correctly.  I think linking to GitHub, as we currently do is less work for me as a maintainer.

This is why I'll close this as won't fix.

PR #374, opened 2019-10-05
stack.yaml: update to latest nightly snapshot

peti:
`stack` version 2.1.3.1 builds the package just fine with this simplified configuration, but the version you're using in CI seems to have trouble with ghc-8.8.1.


sol:
Rebased and pushed to `master`.  Thanks for opening this PR.

Issue #380, opened 2020-03-19
Feature request: add option to omit hash


simfleischman:
## Problem
The generated cabal file has a hash at the top. Example:
```cabal
-- This file has been generated from package.yaml by hpack version 0.31.2.
--
-- see: https://github.com/sol/hpack
--
-- hash: f4d99c3891788b8cb735558ccd259cf1ab897bfdba73efd7266f137e584c7357
```

The hash is useful for checking whether or not to regenerate the cabal file and save directory traversals and the other work hpack does.

However, it forces a **source-control conflict** every time the backed `package.yaml` is updated in separate commits and if the generated cabal file is in source control.

Having the generated file in source control has its tradeoffs, but there are reasons why one would want to do so. One reason is to avoid unnecessary work. Another reason is having consistent builds, which may be affected by having to generate files. I mention these reasons not because everyone should do this, but because there are reasonable reasons to check-in the generated cabal file, and doing so has the effect of frequent source conflicts.

### Add option
Add an option in the `package.yaml` file to omit generating the hash.

Proposed option: `hpack-hash` specified as an optional top-level attribute of type boolean with default value of `true`.

Other choices: `generated-hash` or just `hash`.

### Behavior
In terms of hpack behavior, always regenerate the cabal file seems the most correct course of action. However it is not the most efficient. But perhaps that is the tradeoff of using this feature.

simonmichael:
Yes please! This would absolutely be a more useful default for me when using stack. Regenerating based on modification time is just fine, and I don’t need to know which version of hack it was. This would mean I no longer have to commit cabal file updates separately to avoid conflicts. (And committing cabal files is essential for interop with the non hpack world, and for calming the anti hpack drumbeat in #haskell btw).

simfleischman:
#390 addresses my request.

sol:
As already hinted at https://github.com/sol/hpack/pull/390#issuecomment-626658960, I'm tempted to remove the `hash` feature entirely with `0.35.0`.

PR #381, opened 2020-03-20
Implement hpack-hash feature.

simfleischman:
This implements the feature request in issue #380.

I used the name `hpack-hash` to make it clear that it is a hpack feature, not a cabal one.

I am happy to change the implementation and attribute name as needed.

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -59,15 +59,21 @@ import           Hpack.CabalFile
 programVersion :: Version -> String
 programVersion v = "hpack version " ++ Version.showVersion v
 
-header :: FilePath -> Version -> Hash -> String
-header p v hash = unlines [
+header :: FilePath -> Version -> Maybe Hash -> String
+header p v mHash = unlines ([
     "-- This file has been generated from " ++ takeFileName p ++ " by " ++ programVersion v ++ "."
   , "--"
   , "-- see: https://github.com/sol/hpack"
-  , "--"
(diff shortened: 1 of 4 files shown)


simonmichael:
Could you make it also suppress the hpack version, and anything that causes vcs conflicts. Cf #380

arianvp:
Thanks so much for this! I've been pulling my hairs for hours why i couldn't build a specific `stack.lock.yaml` file anymore only to realise that my OS ships with a different version of `hpack` than `stack` ships with by default https://github.com/NixOS/nixpkgs/issues/84820#issue-597327259

Not having the Hpack version in the generated cabal file seems to exactly solve that problem

sol:
@simfleischman thanks for working on this!

- I agree that hashes in generated cabal files are a pain.  In the past, there were sporadic complaints about `stack` overwriting modified cabal files, this is why @snoyberg  eventually proposed modification detection by using hashes.  This made sense at the time, but in retrospect I'm puzzled whether it hurt more people than it helped.

- From my perspective `package.yaml` is a *definition for a Haskell package*.  With that perspective, I'm not sure if adding `hpack-hash` to `package.yaml` is a good fit.  I opened #390 as an alternative proposal. Everybody who cares about this, please try it and make sure that it solves your use case.

- I understand that the hpack version can lead to friction, but it also serves a purpose.  Let's treat this as a separate story.

simfleischman:
#390 meets this need

PR #390, opened 2020-05-02
Add command-line options --hash and --no-hash

sol:
This is an alternative proposal for #381.  It adds two new command-line flags, `--no-hash` and `--hash`.  A brief description of the semantics follows.

Conceptually, there are four different modes, of which `hpack` exposes three to the user.

### `ForceNoHash`

(enabled with `--no-hash`, implies `--force`)

- No hash will be included in the header comment of the generated cabal file

### `ForceHash`

(enabled with `--hash`, implies `--force`)

- A hash will be included in the header comment of the generated cabal file

### `PreferNoHash`

(the default behavior for `hpack`)

- When there is no existing cabal file it behaves like `ForceNoHash`
- When there is an existing cabal file the convention from that file is followed

### `PreferHash`

(not exposed, but available through the API)

- When there is no existing cabal file it behaves like `ForceHash`
- When there is an existing cabal file the convention from that file is followed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
-## Changes in 0.34.0 (upcoming)
+## Changes in 0.34.0
+  - Use `PreferNoHash` as default `GenerateHashStrategy`
   - Add support for library `visibility` (see #382)
   - Reject URLs for `github`
+
+## Changes in 0.33.1
+  - Add `GenerateHashStrategy`.  The default is `PreferHash` for `0.33.0` and
+    will change to `PreferNoHash` with `0.34.0`. See
+    https://github.com/sol/hpack/pull/390) for details.
+
+  - Add command-line options `--hash` and `--no-hash`
+
+## Changes in 0.33.0.1
+  - Silently ignore missing hash when the cabal file content didn't change at
+    all (for forward compatibility with #390)
 
 ## Changes in 0.33.0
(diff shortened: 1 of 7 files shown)


sol:
@snoyberg what behavior do you think will be appropriate for `stack`?  Given that the hashes are a pain to work with when you put `.cabal` under revision control, I'm tempted to remove support for them entirely.  Would that work for you?

snoyberg:
I honestly don't remember the discussion that led to hashes being included in the first place, or what problems the hashes were intended to solve. I can't think of a way this would break Stack, but I may not be remembering a corner case. Do you remember where we discussed this before?

sol:
@snoyberg took me a while to dig it up,  https://github.com/commercialhaskell/stack/issues/3383#issuecomment-341028378.

snoyberg:
It looks like the original issue the hashes are meant to solve is: user edits cabal file instead of hpack file. There's a pretty clear warning at the top of the generated cabal file that it's autogenerated, and hpack files are probably better understood now than when that original issue opened. I wouldn't advocate for removing the hash, but I'm also not opposed to it. Entirely your call :)

snoyberg:
And thanks for doing the digging on finding that, and pinging me on the topic in the first place

Javran:
For people spending time reading every single line of this PR and yet have no clue how to disable existing hash: just delete your `*.cabal` file and let it regenerate.

sol:
@Javran this may not be obvious, but `hpack --force --no-hash` will do what you want.

PR #394, opened 2020-06-02
Accept subcomponents as dependencies (close #382)

### LINKED ISSUE #382: Add support for visibility: public
Thanks for the super tool!

It would be really nice to have support for multiple public libraries: https://fgaz.me/posts/2019-11-14-cabal-multiple-libraries/

At the moment it seems possible to do with `verbatim: visibility: public` as below, but it seems nasty to have to put these libraries in `internal-libraries`.

```yaml
internal-libraries:
  foo:
    source-dirs: foo
    verbatim:
      visibility: public
    dependencies:
      - base < 4.14
```

sol:


sol:
@Ptival can you please give this a try.

PR #399, opened 2020-07-29
Accept github subdirs more than one level deep

alexfmpe:


src/Hpack/Config.hs
```diff
@@ -588,8 +589,8 @@ instance FromValue GitHub where
   fromValue v = do
     input <- fromValue v
     case map T.unpack $ T.splitOn "/" input of
-      [owner, repo, subdir] -> return $ GitHub owner repo (Just subdir)
-      [owner, repo] -> return $ GitHub owner repo Nothing
+      (owner: _) | owner `elem` ["http:", "https:"] -> fail $ "expected owner/repo or owner/repo/subdir, but encountered url instead: " ++ show input
```
alexfmpe, line 592: Had to add some hacky url detection because otherwise urls were being matched as
```
owner: "https:"
repo: ""
subdir: "github.com/sol/hpack/issues/365"
```

Not sure how to handle this robustly

sol (changes requested):
@alexfmpe Hey! Thanks for working on Hpack + sorry for the late response!

Instead of trying to shoehorn this into `github:`, how about we add more complete support for [`source-repository`](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-section-source-repository-source-repository)?  At this point, I'm not eager to support everything from Cabal, but we can start with `type`, `location` and `subdir` (and possibly extend it in the future).

```yaml
source-repository:
  type: git
  location: https://github.com/sol/hpack
  subdir: some/lengthy/path
```

`source-repository: ...` should take precedence over `github: ...`, that is if both `github: ...` and `source-repository: ...` are specified, then `homepage`, `bug-reports` should be derived from `github: ...` while the `.cabal` `source-repository head` should derived from `source-repository: ...`

Issue #400, opened 2020-10-02
Incorrect cabal-version is generated


gromakovsky:
My `package.yaml` is very small:
```
name:                indigo

library:
  source-dirs: src
  generated-other-modules:
    - Paths_indigo

  default-extensions: [ RebindableSyntax, OverloadedStrings]
```

It produces the following `indigo.cabal` file:
```
cabal-version: 2.0

-- This file has been generated from package.yaml by hpack version 0.34.2.
--
-- see: https://github.com/sol/hpack

name:           indigo
version:        0.0.0
build-type:     Simple

library
  exposed-modules:
      Indigo
  other-modules:
      Paths_indigo
  autogen-modules:
      Paths_indigo
  hs-source-dirs:
      src
  default-extensions: RebindableSyntax OverloadedStrings
  default-language: Haskell2010
```

When I run `cabal build` I get this:
```
cabal: The package uses RebindableSyntax with OverloadedStrings or
OverloadedLists in default-extensions, and also Paths_ autogen module. That
configuration is known to cause compile failures with Cabal < 2.2. To use
these default-extensions with Paths_ autogen module specify at least
'cabal-version: 2.2'.
```

So I suppose `hpack` should set `cabal-version` to 2.2 in this case. As a workaround I do
```
verbatim:
  cabal-version: 2.2
```
but it would be nicer if it was handled by `hpack` itself.

sol:
@gromakovsky hey! Thanks for reporting this.  https://github.com/sol/hpack/pull/405 is my attempt to address this.  Please let me know if you have any feedback.

gromakovsky:
#405 is closed and doesn't have any commits as of now. AFAIU the fix is in https://github.com/sol/hpack/commit/bd27333d7a3c52b14674e51ae13e140e25e7682a. I'll test it when a new release reaches me or maybe will test manually with master. This issue is closed anyway, so there should be no rush. Thanks for the fix!

PR #404, opened 2020-11-04
Omit false conds

sol:


README.md
```diff
@@ -394,6 +412,8 @@ becomes
     else
       ghc-options: -O0
 
+**Note:** Conditionals with `condition: false` are omitted from the generated
+`.cabal` file.
```
sol, line 416: @tfausak do you see any potential issues with this behavior.  One reason why I'm eager to do this is that it will make the fix for https://github.com/sol/hpack/issues/400 less convoluted.

sol: It also gives us cleaner `.cabal` files, which doesn't hurt.

README.md
```diff
@@ -35,6 +35,25 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
 
 ## Documentation
 
+### Handling of `Paths_` modules
+
+Cabal generates a `Paths_` module fore every package.  By default Hpack adds
```
tfausak, line 40:
```suggestion
Cabal generates a `Paths_` module for every package.  By default Hpack adds
```

README.md
```diff
@@ -35,6 +35,25 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
...
+when using a custom `Prelude`).
+
+To prevent Hpack from adding the `Paths_` module to `other-modules` add the
+following to `package.yaml`:
+
+```yaml
+library:
+  when:
+  - condition: false
+    other-modules: Paths_name # substitute name with the package name
```
tfausak, line 54: Should we include a note here about how Cabal generates these modules names? I think it replaces hyphens with underscores. I'm not sure if it does anything else.

README.md
tfausak: Seems safe to me!

src/Hpack/Config.hs
```diff
@@ -468,17 +468,16 @@ hasKey key (Object o) = HashMap.member key o
...
-newtype Cond = Cond String
+data Cond = CondBool Bool | CondExpression String
   deriving (Eq, Show)
 
 instance FromValue Cond where
   fromValue v = case v of
-    String s -> return (Cond $ T.unpack s)
-    Bool True -> return (Cond "true")
-    Bool False -> return (Cond "false")
+    String c -> return (CondExpression $ T.unpack c)
```
tfausak, line 479: I think this means someone could avoid this new filtering behavior by explicitly passing a string. Like this:

``` yaml
when:
  - condition: 'false'
```

That's probably fine, and may in fact be desired as a workaround.

tfausak (approved):
Looks good to me! 

Along the same lines, would it make sense to flatten conditionals where the value is always true?

README.md
sol: That's the only thing I'm aware of.  Ideally, I would like to link to the Cabal docs on `Path_`, however, I couldn't find anything. 

@phadej do you know if the Cabal docs cover `Paths_` anywhere?

src/Hpack/Config.hs
sol: Yes, exactly, that's what I thought.

sol:
> would it make sense to flatten conditionals where the value is always true

Possibly, however, I'm not going to do this for now.

sol:
@tfausak thanks for the review.  I'm going to merge.

PR #408, opened 2020-11-10
Fix #406 by nubbing module list generated from source files

### LINKED ISSUE #353: Handle hsc files more robustly
to reproduce:

- compile an `hsc` file in your project to its `hs` form (in order to make ghci work).
- run `stack --build`.
- hpack will find the `hs` file and add an entry to the module list.

this will lead to two identical entries for the module, one motivated by the `hsc` file, one motivated by the `hs` file.

desired behavior: if there are two files for one module, one `hsc` and one `hs`, only make one entry.

this is arguably not a bug but odd user behavior, but hpack may be in the best position to solve it.  of course i am also very open to ideas how to fix my tooling.  :-)

thanks!

### LINKED ISSUE #406: Duplicate entries in connection with Alex and Happy
This might be the same problem as #353 .
From these files in the source directory:
```
LBNF
├── Abs.hs
├── ErrM.hs
├── Lex.hs
├── Lex.x
├── Par.hs
├── Par.y
├── Print.hs
├── Skel.hs
└── Test.hs
```
I am getting this entry in the generate `.cabal` file:
```
library
  exposed-modules:
      LBNF.Abs
      LBNF.ErrM
      LBNF.Lex
      LBNF.Lex
      LBNF.Par
      LBNF.Par
      LBNF.Print
      LBNF.Skel
      LBNF.Test
```
Cabal does not like this:
```
$ hpack && cabal build
generated Foo.cabal
Resolving dependencies...
Error:
    Module name  LBNF.Lex  is exported multiple times.
    In the stanza 'library'
    In the inplace package 'Foo...'
```
Manually deleting the duplicate entries helps.

Suggestion: `nub` the generated list.

andreasabel:
Fix #406 by nubbing module list generated from source files
    
Inserting a `nub` here is strictly improving the state of affairs, since cabal chokes on a module list with duplicates.
    
This patch might also fix #353.

### DIFF
--- a/src/Hpack/Module.hs
+++ b/src/Hpack/Module.hs
@@ -46,7 +46,7 @@ getModules dir src_ = sortModules <$> do
   if exists
     then do
       src <- Directory.canonicalizePath (dir </> src_)
-      removeSetup src . map toModule <$> getModuleFilesRecursive src
+      removeSetup src . nub . map toModule <$> getModuleFilesRecursive src
     else return []
   where
     removeSetup :: FilePath -> [Module] -> [Module]
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -5,7 +5,7 @@ cabal-version: 1.12
 -- see: https://github.com/sol/hpack
 
 name:           hpack
-version:        0.34.2
+version:        0.34.3
 synopsis:       A modern format for Haskell packages
 description:    See README at <https://github.com/sol/hpack#readme>
 category:       Development
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name: hpack
-version: 0.34.2
+version: 0.34.3
 synopsis: A modern format for Haskell packages
 description: See README at <https://github.com/sol/hpack#readme>
 maintainer: Simon Hengel <sol@typeful.net>


andreasabel:
The travis-ci job `BUILD=tinc CABALVER=2.2 GHCVER=8.4.3` apparently has problems
https://travis-ci.com/github/sol/hpack/jobs/431630342

GHC 8.4.3 crashes when building cabal:
```
Building library for Cabal-3.2.1.0..
[ 55 of 242] Compiling Distribution.CabalSpecVersion ( Distribution/CabalSpecVersion.hs, dist/dist-sandbox-6fba3b3/build/Distribution/CabalSpecVersion.o )

ghc: panic! (the 'impossible' happened)

  (GHC version 8.4.3 for x86_64-unknown-linux):

	Prelude.!!: index too large
```
Maybe try GHC 8.4.4 instead.

```
tinc: callProcess: cabal "install" "--bindir=$prefix/bin/$pkgid" "Cabal-3.2.1.0" "base-compat-0.11.2" ...
... "http-client-tls-0.3.5.3" (exit 1): failed
```

src/Hpack/Module.hs
```diff
@@ -46,7 +46,7 @@ getModules dir src_ = sortModules <$> do
   if exists
     then do
       src <- Directory.canonicalizePath (dir </> src_)
-      removeSetup src . map toModule <$> getModuleFilesRecursive src
+      removeSetup src . nub . map toModule <$> getModuleFilesRecursive src
```
tfausak, line 49: Should this use [`nubOrd`](https://www.stackage.org/haddock/lts-16.20/containers-0.6.2.1/Data-Containers-ListUtils.html#v:nubOrd) instead?

sol: I was considering this as well.  I'm not eager to require `containers >= 0.6` though.  We can add our own inline definition.

andreasabel: I shouldn't have used `nub`, I know, it is evil.
Instead, since `getModules` returns a sorted list, one could have applied `uniq` in the end which just removes consecutive duplicates.
```haskell
uniq = concat . map head . group
```
Or even better, modify `sortModules` to remove duplicates.
```haskell
sortModules = map Module . Set.toAscList . Set.fromList . map unModule
```
If you dare to derive `Ord` for `Module`, then just:
```haskell
sortModules = Set.toAscList . Set.fromList 
```

sol:
@andreasabel I just released `hpack-0.34.3` to Hackage (fyi).

PR #414, opened 2020-11-10
[ fixup #408 ] use uniq instead of nub

andreasabel:
Since the final list of modules is sorted, it is sufficient to throw out consecutive duplicates, in the style of linx command `uniq`.

`nub` is thus unnecessary (and slightly wasteful).

### DIFF
--- a/src/Hpack/Module.hs
+++ b/src/Hpack/Module.hs
@@ -41,18 +41,21 @@ toModule path = case reverse $ Path.components path of
   file : dirs -> Module . intercalate "." . reverse $ dropExtension file : dirs
 
 getModules :: FilePath -> FilePath -> IO [Module]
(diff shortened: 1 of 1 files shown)


sol:
Admittedly, I like the simplicity of your `uniq` implementation + even more so I like the idea of doing `nub` and `sort` in one step via a `Set` (which would require a hand rolled `Ord` instance if we want to keep lexicographical ordering).

But as we already have 2bed823e98f8436ca58b2d1f4a03886df44cad89 I think there is no important technically reason to do anything right now (technically, I think both of your approaches could be faster, but I assume they won't change anything that is measurable).

andreasabel:
>  I think there is no important technically reason to do anything right now

Indeed, this was just cosmetics, improving on my previous fix.

Issue #255, opened 2018-01-19
Inheritance, auto-concatenation and conditionals can lead to surprising order of ghc-options


quasicomputational:
Given this package.yaml:
```
name: test

ghc-options: A
when:
  condition: impl(ghc >= 8.2)
  ghc-options: B

library:
  ghc-options: C
  when:
    condition: impl(ghc >= 8.0)
    ghc-options: D
```
this is what hpack generates for the library section:
```
library
  other-modules:
      Paths_test
  ghc-options: A C
  if impl(ghc >= 8.2)
    ghc-options: B
  if impl(ghc >= 8.0)
    ghc-options: D
```

On GHC 8.2, this will lead to the options being passed to GHC in the order A C B D (i.e., the lexical order in the generated cabal file). I contend that this is not the user's intent, and that A B C D would be better, on the general principle that specific should be able to override general (and hence go later) and C is more specific than B.

cabal doesn't complain about multiple ghc-options fields. This isn't *quite* documented behaviour, but nearly: ("if two conditional sections appear at the same nesting level, properties specified in the latter will come after properties specified in the former")[http://cabal.readthedocs.io/en/latest/developing-packages.html#meaning-of-field-values-when-using-conditionals]; difference being that the A and C options aren't in conditionals. So maybe this would be better to generate:
```
library
  other-modules:
      Paths_test
  ghc-options: A
  if impl(ghc >= 8.2)
    ghc-options: B
  ghc-options: C
  if impl(ghc >= 8.0)
    ghc-options: D
```

quasicomputational:
Another option: abuse `if true` to get the desired ordering. This is also an effective workaround today!

sol:
@quasicomputational can you motivate how this hurts you in practice?

quasicomputational:
Here's a variation of the code that tripped me up:
```
when:
  condition: impl(ghc >= 8.0)
  ghc-options: -Weverything

executable:
  ghc-options: -Wno-missing-import-lists
```

On GHC 8.0 and 8.2, the arguments that get passed to GHC are, in order, "-Wno-missing-import-lists -Weverything". Which means that "-Wmissing-import-lists" is in effect, which surprised me greatly when it happened.

sol:
Would
```yaml
when:
  condition: impl(ghc >= 8.0)
  then:
    ghc-options: -Weverything -Wno-missing-import-lists
  else:
    ghc-options: -Wno-missing-import-lists
```
[...]

PR #416, opened 2020-11-12
Ignore source files that are used as `main` when inferring modules

### LINKED ISSUE #188: Filter main-is of other test-suites from others' other-modules.
With a directory layout (not so uncommon):

```
tests/Doctest.hs
tests/Spec.hs
tests/Spec/FooSpec.hs
```

and `package.yaml`

```yaml
tests:
  spec:
    main:            Spec.hs
    source-dirs:     test
  doctest:
    main:            Doctest.hs
    source-dirs:     test
```

the following `pkg.cabal` is generated (only relevant parts):

```
test-suite doctest
  type: exitcode-stdio-1.0
  main-is: Doctest.hs
  hs-source-dirs:
      test
  other-modules:
      Spec
      Spec.FooSpec

test-suite spec
  type: exitcode-stdio-1.0
  main-is: Spec.hs
  hs-source-dirs:
      test
  other-modules:
      Doctest
      Spec.FooSpec
```

---

This haven't been a problem so-far as `other-modules` weren't yet passed to GHC by `cabal-install` for tests and benchmarks.
Newest cabal will however pass them, which will cause GHC to fail with

```
test/Doctest.hs:1:8: error:
    File name does not match module name:
    Saw: ‘Main’
    Expected: ‘Doctest’
  |
1 | module Main (main) where
```

The reason to pass `other-modules` is for GHC to avoid interleaving module
discovery (cabal-file already tells which modules to compile!) and compilation in GHC.

---

Even proper solution would be to have a way to turn off/configure module discovery per executable/test-suite/benchmark, in addition to filtering out `main-is` files.

sol:
(fixes #188)

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -1096,6 +1096,19 @@ toExecutableMap name executables mExecutable = do
 
 type GlobalOptions = CommonOptions CSources CxxSources JsSources Empty
 
+executableSectionMainCandidates :: Section ExecutableSection -> [FilePath]
+executableSectionMainCandidates sect = case sectionAll sectionSourceDirs sect of
+  [] -> mains
+  sourceDirs -> [src </> main | main <- mains, src <- sourceDirs]
+  where
+    mains :: [FilePath]
+    mains = concatMap sectionMain sect
+
+    sectionMain :: ExecutableSection -> [FilePath]
+    sectionMain exec = case parseMain <$> executableSectionMain exec of
+      Just (file, []) -> [file]
+      _ -> []
+
 toPackage_ :: MonadIO m => FilePath -> Product GlobalOptions (PackageConfig CSources CxxSources JsSources) -> Warnings m (Package, String)
 toPackage_ dir (Product g PackageConfig{..}) = do
   executableMap <- toExecutableMap packageName_ packageConfigExecutables packageConfigExecutable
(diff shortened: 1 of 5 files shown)


sol:
This PR attempts to address https://github.com/sol/hpack/issues/188.  However, this is non-trivial.  I'm not even sure if there is a sound solution.  Consider the following:

```yaml
# source files:
#
#   linux/L.hs
#   windows/W.hs
#   src/L.hs

executable:
  when:
    condition: os(windows)
    then:
      source-dirs:
        - windows
        - src
      main: W.hs
    else:
      source-dirs: linux
      main: L.hs
```
#416 will produce the following output for this:

```cabal
executable foo
  if os(windows)
    main-is: W.hs
    hs-source-dirs:
        windows
        src
  else
    main-is: L.hs
    hs-source-dirs:
        linux
```

specifically, `other-modules: L` is not mentioned under `if os(windows)`, while arguably, a user would want to have it there.

Ignoring any code for now, can we conceptually come up with an algorithm that produces sensible results?

One of the issues is that `main-is` and `hs-source-dirs` "interact" with each other, but they can be specified in different places.  From what I understand, Cabal will evaluate the conditionals, eliminating everything that is `false` and then interpret the whole thing as if it were specified sequentially, adding a kind of procedural component to the mix (akin to what @quasicomputational already described in https://github.com/sol/hpack/issues/255).

 For the example above, we know that the two branches of the `if`-statement are mutually exclusive, giving us more information to work with. But conceptually you could have something like:

```yaml
when:
  - condition: ...
    main: ...
  - condition: ...
    source-dirs: ...
  - condition: ...
    ...
```
Here I don't think it is statically decidable how things interact. (statically == independent of a specific build configuration)

sol:
@mpilgrem @andreasabel do you have any opinion on this?

mpilgrem:
My immediate reaction, which may be harsh, is that people often co-locate related modules but this 'problem' (#188) arises because somebody has co-located unrelated modules (modules that relate to one test suite executable have been co-located with ones that relate to a different test suite executable). Is that a 'problem' that Hpack wants to fix?

EDIT: The Stack project has modules for unit tests that are co-located. However, those modules _are_ related, as follows. They all relate to a _single_ test suite executable, as the Stack project uses the Hspec framework and:
~~~haskell
-- Spec.hs
{-# OPTIONS_GHC -Wno-missing-export-lists #-}
{-# OPTIONS_GHC -F -pgmF hspec-discover   #-}
~~~

andreasabel:
There has not been a lot of reaction on the original issue report from others (no "me too" chime-ins), so maybe the problem is not pressing.

PR #434, opened 2021-02-18
Add support for rts-options

sol:
This PR addresses https://github.com/sol/hpack/issues/237 by adding explicit support for `rts-options`.

**A note on semantics:** When `-with-rtsopts` are specified multiple times then `ghc` only takes the last occurrence into account.  For that reason, when a user wants to add additional rts options within a conditional s/he currently has to reiterate any outer rts options within the conditional, e.g.:

```yaml
executable:
  ghc-options: '"--with-rtsopts -s"'
  when:
    condition: flag(with-threading)
    ghc-options: '"--with-rtsopts -s -N"'
```

I implemented `rts-options` so that they are combined in the same way as most other things we support.  The following gives the same result as above:

```yaml
executable:
  rts-options: -s
  when:
    condition: flag(with-threading)
    rts-options: -N
```

The tests show how this is mapped to `.cabal`.

@snoyberg @quasicomputational @tfausak any thoughts?

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -231,10 +231,11 @@ data ExecutableSection = ExecutableSection {
   executableSectionMain :: Maybe FilePath
 , executableSectionOtherModules :: Maybe (List Module)
 , executableSectionGeneratedOtherModules :: Maybe (List Module)
+, executableSectionRtsOptions :: Maybe (List String)
 } deriving (Eq, Show, Generic, FromValue)
 
 instance Monoid ExecutableSection where
-  mempty = ExecutableSection Nothing Nothing Nothing
+  mempty = ExecutableSection Nothing Nothing Nothing Nothing
   mappend = (<>)
 
 instance Semigroup ExecutableSection where
@@ -242,6 +243,7 @@ instance Semigroup ExecutableSection where
       executableSectionMain = executableSectionMain b <|> executableSectionMain a
     , executableSectionOtherModules = executableSectionOtherModules a <> executableSectionOtherModules b
     , executableSectionGeneratedOtherModules = executableSectionGeneratedOtherModules a <> executableSectionGeneratedOtherModules b
+    , executableSectionRtsOptions = executableSectionRtsOptions a <> executableSectionRtsOptions b
     }
 
 data VerbatimValue =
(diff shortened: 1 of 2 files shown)


tfausak:
Nice! 

The tests only show `rts-options` taking a string. Does it also allow for lists like `ghc-options` does (IIRC)?

src/Hpack/Config.hs
```diff
@@ -231,17 +231,19 @@ data ExecutableSection = ExecutableSection {
   executableSectionMain :: Maybe FilePath
 , executableSectionOtherModules :: Maybe (List Module)
 , executableSectionGeneratedOtherModules :: Maybe (List Module)
+, executableSectionRtsOptions :: Maybe (List String)
```
sol, line 234: @tfausak this specifies the syntax, so yes, it can be a list, a string, or it can be absent.

istathar:
I hope you will consider merging this. It would be a nice addition to **hpack**.

test/EndToEndSpec.hs
```diff
@@ -1436,6 +1436,33 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
 
+      describe "rts-options" $ do
+        it "maps rts-options to ghc-options" $ do
+          [i|
+          executable:
+            main: Main.hs
+            rts-options: -s -N
+          |] `shouldRenderTo` executable_ "foo" [i|
+            main-is: Main.hs
+            ghc-options: "-with-rtsopts -s -N"
```
tfausak, line 1447: Do these need to be quoted, like `-with-rtsopts='-s -N'`? Or perhaps the flag can be duplicated, like `-with-rtsopts=-s -with-rtsopts=-N`.

tfausak (commented):
Oh dear, I hope this hasn't been waiting for a review from me this whole time!

test/EndToEndSpec.hs
sol: From reading the motivation for this PR, this it's what I capture: Giving `--with-rtsopts` multiple times is not accumulative. That's the motivation for this PR in the first place.

Regarding quoting, I assume that what this PR currently does works with Cabal. I haven't  tried it again just yet, but pretty sure that I tried when I originally wrote this.


`--with-rtsopts='-s -N'` may work as well, somebody would need to try. I'm fine with any thing that works here, I'm not very opinionated on this.

sol:
@istathar at the moment I don't have free cycles for this. If you want to help move this forward, this are the things we need:
- [ ] double check that it works as expected and optionally do some bikeshedding on the quoting
- [ ] update the read me
- [ ] add an entry to the change log

As a separate PR:
- [ ] bump the minor version

test/EndToEndSpec.hs
tfausak: This Cabal issue is very related: https://github.com/haskell/cabal/issues/4818

tfausak (approved):

test/EndToEndSpec.hs
sol: Thanks @tfausak for doing all the heavy lifting.

I rarely talk about it in public, but I have had issues with RSI for several years now. when it flares up, I have to stay away from the computer. This is the reason why sometimes I'm less responsive than what I would hope to be.

The TL;DR of it is that I am away from the computer. We still need the  read me update, change  log entry and version bump. I can try to take care of it eventually. But if somebody wants to lend a hand, then that's still very much appreciated.

Releases are fully automated once the version bump hits main.

istathar: I'd be happy to do the _README.md_ update etc today.

istathar:
@sol happy to help with your checklist. I'll see if I can do that this week.

PR #446, opened 2021-07-14
Removed dead constructor PreferHash

andreasabel:
The constructor PreferHash of GenerateHashStrategy is never used,
removing it clarifies the logic to the reader of the code.

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -81,7 +81,10 @@ data Options = Options {
 , optionsToStdout :: Bool
 }
 
-data GenerateHashStrategy = ForceHash | ForceNoHash | PreferHash | PreferNoHash
+data GenerateHashStrategy
+  = ForceHash     -- ^ Option @--hash@ given.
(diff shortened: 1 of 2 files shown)


src/Hpack.hs
```diff
@@ -81,7 +81,10 @@ data Options = Options {
 , optionsToStdout :: Bool
 }
 
-data GenerateHashStrategy = ForceHash | ForceNoHash | PreferHash | PreferNoHash
+data GenerateHashStrategy
+  = ForceHash     -- ^ Option @--hash@ given.
+  | ForceNoHash   -- ^ Option @--no-hash given.
+  | PreferNoHash  -- ^ None of these option given, default behavior.
```
sol, line 87: I think you are right that as far as the `hpack` executable is concerned this is indeed dead code.  However, it is exposed through the API, so as I understand it a user of the API could still use it to force pre-0.34 behavior.

andreasabel: Ok, so maybe phase this out slowly, or deprecate it in the API first.

PR #447, opened 2021-07-14
CI: removed cron job

andreasabel:
This job has been failing for a month without triggering any
maintainance activity.

Save cycles, save CO2!


sol:
I fixed the job instead, closing.

PR #480, opened 2022-03-05
Fix a typo in the name of a type synonym

mpilgrem:


sol:
@mpilgrem nice catch! Thanks a lot!

PR #482, opened 2022-03-08
Fix #481 Allow `default-language` in `package.yaml`

### LINKED ISSUE #481: `default-language` and `GHC2021`
`package.yaml` currently excludes Cabal's `default-language`, because it is taken to be `Haskell2010` by `hpack`. This can be overridden using `verbatim`. With the introduction of `GHC2021` with GHC 9.2.1 (and other possible `GHC20xx` in the future), perhaps `package.yaml` should facilitate `GHC2021` more directly?

mpilgrem:
The default remains as `Haskell2010` but this also allows `GHC2021`.

The documentation is also updated, accordingly.

Tested by compiling and using with GHC 9.2.1 on Windows.

### DIFF
--- a/src/Data/Aeson/Config/FromValue.hs
+++ b/src/Data/Aeson/Config/FromValue.hs
@@ -42,6 +42,7 @@ module Data.Aeson.Config.FromValue (
 
 import           Imports
 
+import           Data.Monoid (Last(..))
 import           GHC.Generics
 
 import           Data.Map.Lazy (Map)
@@ -142,6 +143,9 @@ instance (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 a)) where
 
 instance {-# OVERLAPPING #-} (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 (Maybe a))) where
   genericDecode = accessFieldWith (.:?)
+
+instance {-# OVERLAPPING #-} (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 (Last a))) where
+  genericDecode = accessFieldWith (\ value key -> Last <$> (value .:? key))
 
 accessFieldWith :: forall sel a p. Selector sel => (Object -> Key -> Parser a) -> Options -> Value -> Parser (S1 sel (Rec0 a) p)
 accessFieldWith op Options{..} v = M1 . K1 <$> withObject (`op` Key.fromString label) v
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -37,6 +37,7 @@ module Hpack.Config (
 , packageDependencies
 , package
 , section
+, sectionWithHaskell2010
 , Package(..)
 , Dependencies(..)
 , DependencyInfo(..)
@@ -59,6 +60,7 @@ module Hpack.Config (
 , Cond(..)
 , Flag(..)
 , SourceRepository(..)
+, Language(..)
 , BuildType(..)
 , GhcProfOption
 , GhcjsOption
@@ -88,6 +90,7 @@ import           Data.Map.Lazy (Map)
 import qualified Data.Map.Lazy as Map
 import qualified Data.Aeson.Config.KeyMap as KeyMap
 import           Data.Maybe
+import           Data.Monoid (Last(..))
 import           Data.Ord
 import qualified Data.Text as T
 import           Data.Text.Encoding (decodeUtf8)
@@ -185,7 +188,10 @@ packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . f
     deps xs = [(name, info) | (name, info) <- (Map.toList . unDependencies . sectionDependencies) xs]
 
 section :: a -> Section a
-section a = Section a [] mempty [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
+section a = Section a [] mempty [] [] [] Nothing [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
+
+sectionWithHaskell2010 :: a -> Section a
+sectionWithHaskell2010 a = (section a) {sectionDefaultLanguage = Just $ Language "Haskell2010"}
 
 packageConfig :: FilePath
 packageConfig = "package.yaml"
@@ -271,6 +277,7 @@ data CommonOptions cSources cxxSources jsSources a = CommonOptions {
 , commonOptionsPkgConfigDependencies :: Maybe (List String)
 , commonOptionsDefaultExtensions :: Maybe (List String)
 , commonOptionsOtherExtensions :: Maybe (List String)
+, commonOptionsDefaultLanguage :: Last Language
 , commonOptionsGhcOptions :: Maybe (List GhcOption)
 , commonOptionsGhcProfOptions :: Maybe (List GhcProfOption)
 , commonOptionsGhcjsOptions :: Maybe (List GhcjsOption)
@@ -304,6 +311,7 @@ instance (Semigroup cSources, Semigroup cxxSources, Semigroup jsSources, Monoid 
   , commonOptionsPkgConfigDependencies = Nothing
   , commonOptionsDefaultExtensions = Nothing
   , commonOptionsOtherExtensions = Nothing
+  , commonOptionsDefaultLanguage = Last Nothing
   , commonOptionsGhcOptions = Nothing
   , commonOptionsGhcProfOptions = Nothing
   , commonOptionsGhcjsOptions = Nothing
@@ -328,6 +336,9 @@ instance (Semigroup cSources, Semigroup cxxSources, Semigroup jsSources, Monoid 
   }
   mappend = (<>)
 
+defaultLanguage :: Language
+defaultLanguage = Language "Haskell2010"
+
 instance (Semigroup cSources, Semigroup cxxSources, Semigroup jsSources) => Semigroup (CommonOptions cSources cxxSources jsSources a) where
   a <> b = CommonOptions {
     commonOptionsSourceDirs = commonOptionsSourceDirs a <> commonOptionsSourceDirs b
@@ -335,6 +346,7 @@ instance (Semigroup cSources, Semigroup cxxSources, Semigroup jsSources) => Semi
   , commonOptionsPkgConfigDependencies = commonOptionsPkgConfigDependencies a <> commonOptionsPkgConfigDependencies b
   , commonOptionsDefaultExtensions = commonOptionsDefaultExtensions a <> commonOptionsDefaultExtensions b
   , commonOptionsOtherExtensions = commonOptionsOtherExtensions a <> commonOptionsOtherExtensions b
+  , commonOptionsDefaultLanguage = commonOptionsDefaultLanguage a <> commonOptionsDefaultLanguage b
   , commonOptionsGhcOptions = commonOptionsGhcOptions a <> commonOptionsGhcOptions b
   , commonOptionsGhcProfOptions = commonOptionsGhcProfOptions a <> commonOptionsGhcProfOptions b
   , commonOptionsGhcjsOptions = commonOptionsGhcjsOptions a <> commonOptionsGhcjsOptions b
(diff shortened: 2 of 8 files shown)


sol:
@mpilgrem I'll try to look at this some time soon.  For the meantime you can use `verbatim` to override `default-language`.

mpilgrem:
Looking at the CI failures, something is going wrong with the conditions. I'll have to revisit the pull request.

mpilgrem:
I changed my approach which, I think, fixed the problem with the conditionals. I also modified some of the existing tests to specify the section default language as `Haskell2010` in the 'should be' comparisons, which causes the tests to pass (appropriately, I think).

mpilgrem:
I have realised that I should not have used `mempty` to introduce the default value of `GhcLanguage`. I will need to revisit the pull request again.

mpilgrem:
To explain further my proposed solution:

In module `Hpack.Config`, I introduce type `GhcLanguage` with constructors `Haskell2010` and `GHC2021`, and make it an instance of `FromValue`. It is exported (as it is needed in module `Hpack.Render`). `defaultLanguage` is equated with `Haskell2010`.

I add field `commonOptionsDefaultLanguage :: Last GhcLanguage` to constructor `CommonOptions` (importing `Data.Monoid (Last (..))`).

`defaultCommonOptions` is equated with `mempty` but with `commonOptionsDefaultLanguage = Last (Just defaultLanguage)`.

In `readPackageConfig`, the use of `config` is replaced by `configWithDefaultCommonOptions`, which is `config` updated for `defaultCommonOptions` (using `<>`, hence the use of `Data.Monoid.Last`).

I add field `sectionDefaultLanguage :: Maybe GhcLanguage` to constructor `Section`, and `toSection` equates `sectionDefaultLanguage = getLast commonOptionsDefaultLanguage`.

In module `Data.Aeson.Config.FromValue`, `S1 sel (Rec0 (Last a))` is added as an instance of `GenericDecode`, following the existing instance for `S1 sel (Rec0 (Maybe a))`. A helper operator `.*` is defined to assist with the new instance.

In module `Hpack.Render`, `defaultLanguage` is no longer in the `extraFields` lists and `defaultLanguage` is no longer required.

`renderDefaultLanguage :: GhcLanguage -> Element` is added.

`renderSection` adds `++ maybe [] (return . renderDefaultLanguage) sectionDefaultLanguage` to the end.

src/Hpack/Config.hs
```diff
@@ -501,6 +509,20 @@ instance Semigroup Empty where
 instance FromValue Empty where
   fromValue _ = return Empty
 
+data GhcLanguage =
+    Haskell2010
```
andreasabel, line 513: Why not also allow `Haskell98`?

andreasabel (commented):
This feature would be very much appreciated!!

src/Hpack/Config.hs
mpilgrem: I don't think there is any reason to exclude Haskell98. I only left it out because the existing `hpack` does not accommodate it.

andreasabel: I think it should be added.  Otherwise the `verbatim: default-language: Haskell98` hack will prevail.

mpilgrem: Now added.

sol:
A better workaround is to use `default-extensions`:

```yaml
default-extensions: GHC2021
```
You will still have `default-language: Haskell2010` in the generated `.cabal` file.  But `GHC2021` will take precedence.

src/Data/Aeson/Config/FromValue.hs
```diff
@@ -143,6 +147,9 @@ instance (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 a)) where
 instance {-# OVERLAPPING #-} (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 (Maybe a))) where
   genericDecode = accessFieldWith (.:?)
 
+instance {-# OVERLAPPING #-} (Selector sel, FromValue a) => GenericDecode (S1 sel (Rec0 (Last a))) where
+  genericDecode = accessFieldWith (fmap Last .* (.:?))
```
sol, line 151: I think
```haskell
  genericDecode = accessFieldWith (\ value key -> Last <$> (value .:? key))
```
is easier to understand.

I like the use of `Last`, but if you want to add the instance you also need to add tests (similar to what I just did for `Maybe` in https://github.com/sol/hpack/commit/35902395c0dbe2a96b678eef2f93168cd39b80d2, don't add the nested instance, unless we are going to use it though).

Alternatively use `<|>` (as we currently do in other places).

src/Data/Aeson/Config/FromValue.hs
```diff
@@ -67,6 +68,9 @@ decodeValue = runParser fromValue
 (.:?) :: FromValue a => Object -> Key -> Parser (Maybe a)
 (.:?) = explicitParseFieldMaybe fromValue
 
+(.*) :: (c -> d) -> (a -> b -> c) -> (a -> b -> d)
+(f .* g) x y = f (g x y)
```
sol, line 72: :-1:

src/Hpack/Config.hs
```diff
@@ -185,7 +187,7 @@ packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . f
     deps xs = [(name, info) | (name, info) <- (Map.toList . unDependencies . sectionDependencies) xs]
 
 section :: a -> Section a
-section a = Section a [] mempty [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
+section a = Section a [] mempty [] [] [] Nothing [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
```
sol, line 190: 👍

src/Hpack/Config.hs
```diff
@@ -501,6 +509,21 @@ instance Semigroup Empty where
 instance FromValue Empty where
   fromValue _ = return Empty
 
+data GhcLanguage =
+    Haskell98
+  | Haskell2010
+  | GHC2021
+  deriving (Eq, Show, Generic, Enum, Bounded)
```
sol, line 516: What about using a `String` instead and leaving anything further to `Cabal`?   That way older versions of `hpack` will continue to work with the next set of `GHC202x` extensions.

A `newtype` will still be neat.  Could we call it `Language` instead of `GhcLanguage`, or is that already taken?

src/Hpack/Config.hs
```diff
@@ -501,6 +509,21 @@ instance Semigroup Empty where
...
+
+instance FromValue GhcLanguage where
+  fromValue = withText $ \ (T.unpack -> t) -> do
+    maybe err return (lookup t options)
+    where
+      err = fail ("expected one of " ++ formatOrList ghcLanguagesAsString)
+      ghcLanguages = [minBound .. maxBound]
+      ghcLanguagesAsString = map show ghcLanguages
+      options = zip ghcLanguagesAsString ghcLanguages
```
sol, line 525: 👎 copy & paste (if we were to keep it, it needs to get DRYed; but see above)

src/Hpack/Config.hs
```diff
@@ -652,9 +687,13 @@ readPackageConfig (DecodeOptions programName file mUserDataDir readValue) = runE
   (warnings, value) <- lift . ExceptT $ readValue file
   tell warnings
   config <- decodeValue programName file value
+  let Product (Product _ configCommonOptions) _ = config
+      commonOptionsWithDefaults = defaultCommonOptions <> configCommonOptions
+      configWithDefaultCommonOptions =
+        first (second (const commonOptionsWithDefaults)) config
```
sol, line 693: Thus feels a little bit dirty, but I don't have any better idea. I think making it more explicit could help (name it clearly after what's going on here, e.g. `configWithDefaultLanguage` etc, don't abstract early).

src/Hpack/Render.hs
```diff
@@ -186,7 +186,7 @@ renderBenchmark (name, sect) =
     (renderExecutableSection [Field "type" "exitcode-stdio-1.0"] sect)
 
 renderExecutableSection :: [Element] -> Section Executable -> [Element]
-renderExecutableSection extraFields = renderSection renderExecutableFields extraFields [defaultLanguage]
+renderExecutableSection extraFields = renderSection renderExecutableFields extraFields []
```
sol, line 189: I think it's now possible to simplify `renderSection` (remove the last argument). Or isn't it?

src/Hpack/Render.hs
```diff
@@ -252,6 +252,7 @@ renderSection renderSectionData extraFieldsStart extraFieldsEnd Section{..} = ad
   ++ maybe [] (return . renderBuildable) sectionBuildable
   ++ map (renderConditional renderSectionData) sectionConditionals
   ++ extraFieldsEnd
+  ++ maybe [] (return . renderDefaultLanguage) sectionDefaultLanguage
```
sol, line 255: Putting this at the end will prevent you from overriding it in a conditional.  Conditionals have to go last.

test/Hpack/ConfigSpec.hs
```diff
@@ -371,9 +371,9 @@ spec = do
             ghcjs-options: -ghcjs3
         |]
         (`shouldBe` package {
-          packageLibrary = Just (section library) {sectionGhcjsOptions = ["-dedupe", "-ghcjs1"]}
-        , packageExecutables = Map.fromList [("foo", (section $ executable "Main.hs") {sectionGhcjsOptions = ["-dedupe", "-ghcjs2"]})]
-        , packageTests = Map.fromList [("spec", (section $ executable "Spec.hs") {sectionGhcjsOptions = ["-dedupe", "-ghcjs3"]})]
+          packageLibrary = Just (section library) {sectionGhcjsOptions = ["-dedupe", "-ghcjs1"], sectionDefaultLanguage = Just Haskell2010}
+        , packageExecutables = Map.fromList [("foo", (section $ executable "Main.hs") {sectionGhcjsOptions = ["-dedupe", "-ghcjs2"], sectionDefaultLanguage = Just Haskell2010})]
+        , packageTests = Map.fromList [("spec", (section $ executable "Spec.hs") {sectionGhcjsOptions = ["-dedupe", "-ghcjs3"], sectionDefaultLanguage = Just Haskell2010})]
```
sol, line 376: Is it possible to DRY this up?

sol (changes requested):
Thanks for opening this PR and sorry for the late response.

This mostly looks great.

I left a couple of comments.

Please add acceptance tests to `test/EndToEndSpec.hs`.

sol:
@mpilgrem please don't take any of my comments the wrong way.  It's purely technical.  In general, this project follows TDD.  If you are not familiar with TDD then please let me know and I'll help with the tests.

src/Hpack/Render.hs
mpilgrem: It is possible to simplify, and I will do that.

src/Hpack/Config.hs
mpilgrem: I will change it to a `newtype` of a `String`, and name the type/constructor `Language`.

src/Hpack/Config.hs
mpilgrem: This instance definition now falls away, due to the change in the type.

test/Hpack/ConfigSpec.hs
mpilgrem: It is and I will do so.

src/Data/Aeson/Config/FromValue.hs
mpilgrem: Noted. I will also retain `Last` and add tests.

mpilgrem:
@sol, thank you for the review. I think I have made changes to the pull request that are responsive to all of your comments.

test/EndToEndSpec.hs
```diff
@@ -1825,24 +1861,38 @@ library #{name}
   default-language: Haskell2010
 |]
 
+-- As for executable_, but without the final "  default-language: Haskell2020"
+executable'_ :: String -> String -> Package
```
sol, line 1865: This is a difficult decision. On the one hand we want to keep things DRY (like you did 👍), on the other hand we don't want too many versions of ` executable*` and `library*`. We can't have both.

In this particular case I'm in favor of the duplication. I'll take care of it.

sol (approved):
This looks great now 👍

I left a couple of notes. I'll merge and take care of them in a follow up.

sol:
https://github.com/sol/hpack/pull/501

PR #487, opened 2022-03-26
Support Cabal 3.6 (fixes #484)

### LINKED ISSUE #484: Support Cabal library 3.6
`Cabal-3.6` will be required once stackage nightly switches to GHC 9.2: https://github.com/commercialhaskell/stackage/pull/6318/files#diff-9949098b9af80e5fcef9a718fdda416f3fa8272ca8add4d67ae65fe892990939R6462

bergmark:
Attempt to resolve #484. I'm not sure if this is a proper fix?

Side-note, when forking this repo on github, I instead got this... any idea why?

Screenshot 2022-03-26 at 16 09 24

### DIFF
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -1343,7 +1343,7 @@ inferModules ::
   -> IO (Section b)
 inferModules dir packageName_ getMentionedModules getInferredModules fromData fromConditionals = fmap removeConditionalsThatAreAlwaysFalse . traverseSectionAndConditionals
   (fromConfigSection fromData [pathsModuleFromPackageName packageName_])
-  (fromConfigSection (\ [] -> fromConditionals) [])
+  (fromConfigSection (\ _ -> fromConditionals) [])
   []
   where
     fromConfigSection fromConfig pathsModule_ outerModules sect@Section{sectionData = conf} = do
--- a/hpack.cabal
+++ b/hpack.cabal
@@ -27,7 +27,7 @@ library
       src
   ghc-options: -Wall
   build-depends:
-      Cabal >=3.0.0.0 && <3.6
+      Cabal >=3.0.0.0 && <3.7
     , Glob >=0.9.0
     , aeson >=1.4.3.0
     , base >=4.9 && <5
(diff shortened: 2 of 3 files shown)


andreasabel:
> Side-note, when forking this repo on github, I instead got this... any idea why?

Likely because you forked from https://github.com/yamadapc/hpack-convert which forked from _this_, and github seems to allow only one fork per subject/organization.

src/Hpack/Config.hs
```diff
@@ -1343,7 +1343,7 @@ inferModules ::
   -> IO (Section b)
 inferModules dir packageName_ getMentionedModules getInferredModules fromData fromConditionals = fmap removeConditionalsThatAreAlwaysFalse . traverseSectionAndConditionals
   (fromConfigSection fromData [pathsModuleFromPackageName packageName_])
-  (fromConfigSection (\ [] -> fromConditionals) [])
+  (fromConfigSection (\ _ -> fromConditionals) [])
```
sol, line 1346: @bergmark do you know for what reason this change was necessary?  Did you get a test failure?

sol: The code is somewhat involved, but I think it's statically guaranteed that this list is always empty.  Looking at the relevant parts of `fromConfigSection`:

```haskell
  where
    fromConfigSection fromConfig pathsModule_ ... = do
      ...
      let
        ...
        pathsModule = (pathsModule_ \\ outerModules) \\ mentionedModules
        r = fromConfig pathsModule inferableModules conf
      ...
```
Substituting `fromConfig` with `(\ [] -> fromConditionals)` and `pathsModule_` with `[]`:
```haskell
      ...
      let
        ...
        pathsModule = ([] \\ outerModules) \\ mentionedModules
        r = (\ [] -> fromConditionals) pathsModule inferableModules conf
      ...
```

=>
```haskell
      ...
      let
        ...
        pathsModule = []
        r = (\ [] -> fromConditionals) pathsModule inferableModules conf
      ...
```
=>
```haskell
      ...
      let
        ...
        r = (\ [] -> fromConditionals) [] inferableModules conf
      ...
```
we see that the lambda is only ever called with an empty list.

sol:
@bergmark I took the liberty to amend your commit in https://github.com/sol/hpack/pull/489.

src/Hpack/Config.hs
sol: Umm, this is actually because of `-Wincomplete-uni-patterns` being added to `-Wall`.   I'm still going to disable that warning instead, as matching on `[]` communicates that static guarantee to the reader.

bergmark:
Thanks!

PR #494, opened 2022-03-28
Add failing test cases for verbatim having cabal-version.

philderbeast:


test/EndToEndSpec.hs
```diff
@@ -1680,6 +1680,32 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+          |] `shouldRenderTo` (package "") {packageCabalVersion = "3.0"}
+
+        it "cabal-version can be sandwiched between other verbatim fields" $ do
+          [i|
+          verbatim: |
+            common set-lang-2010
+              default-language: Haskell2010
+            cabal-version: 3.0
+            common warn-all
+              ghc-options: -Wall
```
sol, line 1706: You really want to use `defaults` and/or YAML anchors instead.

PR #497, opened 2022-04-21
Update .ghci

sol:


.ghci
```diff
@@ -1 +1 @@
-:set -isrc -itest -idist/build/autogen -DTEST -optP-include -optPdist/build/autogen/cabal_macros.h
+:set -XHaskell2010 -fno-warn-incomplete-uni-patterns -DTEST -isrc -itest -i./dist-newstyle/build/x86_64-linux/ghc-9.2.2/hpack-0.34.7/build/autogen/
```
sol, line 1: Is there a better way to do this with `cabal new-*`?

sol:
As we include `src` in the test suite, we can use `cabal repl` or `cabal repl spec --with-ghc=sensei` for that matter.

PR #504, opened 2022-04-22
Rename `default-language` to `language`

sol:
(accept `default-language` as an alias)


sol:
@mpilgrem @andreasabel I'm going to rename `default-language` to `language`.  `default-language` is still accepted as an alias.

PR #505, opened 2022-04-22
Allow to set `language` from `defaults`

sol:


test/EndToEndSpec.hs
```diff
@@ -982,6 +982,20 @@ spec = around_ (inTempDirectoryNamed "foo") $ do
...
+        |]
+
+        [i|
+        defaults: sol/hpack-template@2017
+        library: {}
+        |] `shouldRenderTo` library [i|
+        other-modules:
+            Paths_foo
+        default-language: GHC2021
+        |]
```
sol, line 997: @mpilgrem there was a bug I fixed.  This didn't work before.

Issue #510, opened 2022-05-08
Enhancement: Allow file exclusion where  file globs are supported


TheDrawingCoder-Gamer:
When adding extra source files, I noticed that it includes files I don't want. However, to get around this I would have to write much more patterns then are needed. Allowing glob pattern exclusion (similar to `!file` syntax in .gitignore files) would make excluding files much easier. 
So e.g.
```yaml
extra-lib-dirs:
- c/**
- !c/pamixer/build/**
```
would include all files under `c` except c/pamixer/build

PR #511, opened 2022-05-08
Allow Exclusion globs

TheDrawingCoder-Gamer:
- [x] Did you add a new feature to `hpack`?
  - [ ] Did you add an acceptance test for that new feature to `test/EndToEndSpec.hs`?
Tested on my project - works well. Need to add testing, then it's ready to merge.


TheDrawingCoder-Gamer:
Fixes #510

PR #519, opened 2022-08-02
Remove broken link in `README.md`

abooij:
It seems like the trunk branch got renamed.

- [ ] Did you add a new feature to `hpack`?
  - [ ] Did you add an acceptance test for that new feature to `test/EndToEndSpec.hs`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -714,10 +714,5 @@ steps are required.
 
 ## Binaries for use on Travis CI
 
-You can get binaries for use on Travis CI with:
-
-```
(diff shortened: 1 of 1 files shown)


sol:
This change makes sense. However, we currently don't produce binary distributions. So this is something we need to do to make this change meaningful.

abooij:
@sol How about just removing the link, then? I've updated the PR.

mpilgrem:
@abooij, I made a change to put 'currently' in the context of 'previously'. I also tidied-up the commit history, in preparation for merging.

PR #520, opened 2022-08-08
Add examples for flags in README

futtetennista:
I couldn't find any documentation of flags and I ended up looking at the code. It wasn't clear to me from the current docs:

1. what was the top-level object name
2. wether individual flags had to be specified as objects or a list

- [x] Did you add a new feature to `hpack`? No
- [x] Did you add an acceptance test for that new feature to `test/EndToEndSpec.hs`? Not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,20 @@ This is done to allow compatibility with a wider range of `Cabal` versions.
 | `manual` | · | | Required (unlike Cabal) |
 | `default` | · | | Required (unlike Cabal) |
 
+Flags must be specified as an object:
+
+```
+flags:
+  development:
+    description: A development or production build 
+    manual: True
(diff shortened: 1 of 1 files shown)


ad-si:
@sol Can we please merge this? I was exactly looking for such an example snippet in the readme just now.

mpilgrem (commented):
If an example would be helpful to remove that perceived ambiguity, would it not be better to add it to line 151, so that line is something like (given the limitations of tables in Markdown):
~~~md
| `flags`  | `flag <name>` | | Map from flag name to flag (see [Flags](#flags)) | `flags: { my-cabal-flag: { description: 'Optional description', manual: true, default: false } }` | |
~~~

PR #533, opened 2022-11-26
Add minimal example to README

roosemberth:
## Motivation for this MR

Every time I start working in a new Haskell project, I find myself in this
repository's README, copy one of the provided examples, but spend most of the
time removing keys I don't need. I think a simple example one can simply
'copy over' is both useful to making the tool accessible to beginners and
a practical base for experienced developers to build on.

## Changes in this MR

This MR adds a short and simple example, matching the default `cabal init`
'Library and Executable' generation (`app`, `src`, `test` directories with
`Main.hs` inside them).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,35 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
 
 ## Examples
 
+The following `package.yaml` describes a project with a library under `src`, an executable under `app/Main.hs` and tests under `test/Main.hs`:
+
+```yaml
+name: mylib
+
+dependencies:
+  - base >= 4.9 && < 5
+
+library:
+  source-dirs: src
+
+executable:
+  main: Main.hs
+  source-dirs: app
+  dependencies:
+    - mylib
+
+tests:
+  spec:
+    main: Main.hs
+    source-dirs:
+      - test
+      - src
+    dependencies:
+      - QuickCheck
+```
+
+### More examples
+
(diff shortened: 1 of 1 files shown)


roosemberth:
@sol Gentle ping ^^

README.md
```diff
@@ -32,6 +32,35 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
...
+    - mylib
+
+tests:
+  spec:
+    main: Main.hs
+    source-dirs:
+      - test
+      - src
+    dependencies:
+      - QuickCheck
```
mpilgrem, line 59: For a minimal example, I would expect a dependency on the package itself but not `QuickCheck`.

README.md
```diff
@@ -32,6 +32,35 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
 
 ## Examples
 
+The following `package.yaml` describes a project with a library under `src`, an executable under `app/Main.hs` and tests under `test/Main.hs`:
```
mpilgrem, line 35: I think, strictly, (if you follow my other suggestions below):
~~~md
... a package with a main library with source code under directory `src`, an
executable named `my-exe` with source code under `app`, and a test suite named
`my-test-suite` with source code under `test`:
~~~

README.md
```diff
@@ -32,6 +32,35 @@ at the Singapore Haskell meetup: http://typeful.net/talks/hpack
 
 ## Examples
```
mpilgrem, line 34: It may be worth acknowledging in the introduction that users of Stack have access to templates for projects that include `package.yaml` files, but users of Cabal (the tool) may find the minimal example useful.

PR #551, opened 2023-07-02
Use `crypton` instead of `cryptonite`

sol:


.github/workflows/build.yml
```diff
@@ -37,7 +34,7 @@ jobs:
           - os: macos-latest
             ghc: system
           - os: windows-latest
-            ghc: system
+            ghc: '9.2'
```
sol, line 37: @mpilgrem I think I remember that you opened a GHC issue for the windows build failure. What's the related issue number?

PR #555, opened 2023-08-03
Don't write out the cabal file if it matches the old one

Gabriella439:
The motivation behind this change is avoid the `--force` option from triggering an unnecessary reload of `ghcid` if nothing changed.

The context is that at work we use `hpack --force` to correct manual edits that users might make to the `.cabal` file, and we run `hpack --force` whenever users enter our Nix shell.  If a user uses `direnv` on top of that then `hpack --force` runs every time they `cd` to the project directory.

However, suppose that they load `ghcid` in one window and then `cd` to the project directory in another window.  This causes `hpack --force` to run and even if nothing changed the `mwb.cabal` file is touched which causes `ghcid` to reload (which is expensive).  This change fixes that.

### DIFF
--- a/src/Hpack.hs
+++ b/src/Hpack.hs
@@ -248,7 +248,7 @@ hpackResultWithVersion v (Options options force generateHashStrategy toStdout) =
           NoForce -> maybe Generated (mkStatus newCabalFile) mExistingCabalFile
 
       case status of
-        Generated -> writeCabalFile options toStdout cabalFileName newCabalFile
+        Generated -> writeCabalFile options toStdout cabalFileName mExistingCabalFile newCabalFile
         _ -> return ()
 
       return $ Right Result {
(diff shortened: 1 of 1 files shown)


Gabriella439:
Note that I could also structure this to check if the new/old cabal files are textually identical instead of the abstract syntax being identical.  However, that would be a slightly more invasive change so I wanted to propose this first to solicit feedback.

sol:
> we run `hpack --force` whenever users enter our Nix shell.

What is the reason that you use `--force` in the first place? If you can remove the `--force` option then I think things would already work for your use case.

Do you have a modification hash in the degenerated `.cabal` file? If yes, then run `hpack --no-hash --force` once to get rid of it. After that `hpack` should work just fine, even with manual modifications and without `--force`.

Or is it, that you use different versions of `hpack` across the team?

> Note that I could also structure this to check if the new/old cabal files are textually identical instead of the abstract syntax being identical.

Yes, if we do anything here, then we would want textual comparison. But please hold back for a second.  This code is already more complicated than it should be. We might want to remove support for modification hashes (and possibly even the `hpack` min version check) first, before doing anything here.

sol:
@Gabriella439 please take a look at https://github.com/sol/hpack/pull/557.  Does this work for you?

Gabriella439:
Yeah, #557 looks good to me.  To answer your prior question, we would use `--force` so that if the user made any manual modification to the `.cabal` file then we'd correct and override that modification.

sol:
- #560

PR #564, opened 2023-09-18
Don't infer `Paths_`-module with `spec-version: 0.36.0` or later

sol:


mpilgrem:
@sol, I see that this pull request is still a work in progress. It occured to me that it might be useful to provide here a link to, or documentation of, the motivation for this change in behaviour. The existing documentation states that the existing behaviour is sometimes useful and most of the time not harmful. Has something happened to change the 'useful v harmful' trade-off?

sol:
> Has something happened to change the 'useful v harmful' trade-off?

The issue here arises when you use a custom prelude that doesn't provide what the `Paths_` module expects (e.g. both abstract `FilePath` and abstract `String` won't work, neither individually, nor in combination).

In that situation, you will end up with a compile error for the `Paths_` module, even though you never requested it. This is somehow not very nice.

In retrospect, inferring the `Paths_` module in the first place is a historic mistake.

Requesting it explicit is better here.

The reason why I added inference for the `Paths_` module originally was that back then we didn't have `generated-other-modules`, and it wasn't anything I had imagined back then (this was still pre Cabal 2.0). So the only good option back then was to specify the `Paths_` module as `other-modules`, which then in many cases effectively disabled module inference completely, which was also not very nice.

Custom preludes weren't a thing back then, so inferring the `Paths_` module was the obvious solution.

Now we are stuck with it, and changing the default behavior would break existing packages.

That's why it's still like this today. It's also why I think we need to make the "modern behavior" opt-in (this PR), even though I don't really like it. I would rather have it as the default.

sol:
@mpilgrem provided there are no concerns from your side, I'm going to merge this.

mpilgrem:
No concerns from me. If this is released soon, I'll likely include it in the release of Stack 2.13.1, even though it was not in the release candidate.

PR #567, opened 2023-09-19
Desugar local build tools to `build-tool-depends` (fixes #516)

### LINKED ISSUE #516: build-tools for executable in same package incorrect
hpack `0.34.4`

Currently on this commit: https://github.com/brandonchinn178/tasty-autocollect/tree/1f45427cc4784c94679a14d0ceda72024491ae10

`package.yaml` contains
```yaml
build-tools:
  - tasty-autocollect:tasty-autocollect
```

The cabal file contains
```cabal
build-tools:
    tasty-autocollect
```

Doing a `stack build` from scratch shows the message:

```
No packages found in snapshot which provide a "tasty-autocollect" executable, which is a build-tool dependency of tasty-autocollect
```
But doing a manual
```yaml
verbatim:
  build-tool-depends: tasty-autocollect:tasty-autocollect
```
then building from scratch (`stack purge && stack build`) no longer shows the message

sol:
- This has the potential to break packages for `cabal v1-`, and Hackage still uses `cabal v1-`: https://github.com/haskell/hackage-server/issues/821
- At first glance you might think that cabal files that use `build-tool-depends` should specify at least `cabal-version: 2.0`, however, that's not strictly true, see: [`build-tools`](https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-build-tools) and [`build-tool-depends`](https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-build-tool-depends)

### DIFF
--- a/src/Hpack/Render.hs
+++ b/src/Hpack/Render.hs
@@ -351,7 +351,6 @@ data RenderBuildTool = BuildTools String | BuildToolDepends String
 
 renderBuildTool :: (BuildTool,  DependencyVersion) -> RenderBuildTool
 renderBuildTool (buildTool, renderVersion -> version) = case buildTool of
-  LocalBuildTool executable -> BuildTools (executable ++ version)
   BuildTool pkg executable
     | pkg == executable && executable `elem` knownBuildTools -> BuildTools (executable ++ version)
     | otherwise -> BuildToolDepends (pkg ++ ":" ++ executable ++ version)
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -983,7 +983,7 @@ data Executable = Executable {
 , executableGeneratedModules :: [Module]
 } deriving (Eq, Show)
 
-data BuildTool = BuildTool String String | LocalBuildTool String
+data BuildTool = BuildTool String String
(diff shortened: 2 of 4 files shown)


src/Hpack/Render.hs
```diff
@@ -351,7 +351,6 @@ data RenderBuildTool = BuildTools String | BuildToolDepends String
 
 renderBuildTool :: (BuildTool,  DependencyVersion) -> RenderBuildTool
 renderBuildTool (buildTool, renderVersion -> version) = case buildTool of
```
mpilgrem, line 353: I assume it is a stylistic choice, but I noted the continued use of `case` even though type `BuildTool` has only one constructor.

mpilgrem (approved):
Looks good to me, but does `Hpack.Config.determineCabalVersion` need to be amended if `build-tool-depends` in a Cabal file implies, at least, `cabal-version: 2.0`?

sol:
@mpilgrem thanks a lot for the review. This is not going to land for now. I updated the PR description.

sol:
Closing in favor of https://github.com/sol/hpack/pull/599

PR #570, opened 2023-11-16
Hide Data.List.List import

Kleidukos:
This PR hides the `Data.List.List` identifier from the Data.List import in Imports.hs.

Since the identifier will arrive with [base-4.20.0.0](https://gitlab.haskell.org/ghc/ghc/-/blob/2776920e642544477a38d0ed9205d4f0b48a782e/libraries/base/changelog.md#42000-tba), this hiding is guarded behind a CPP conditional.

### DIFF
--- a/src/Imports.hs
+++ b/src/Imports.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 module Imports (module Imports) where
 
 import           Control.Applicative as Imports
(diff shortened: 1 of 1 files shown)


sol:
Thanks @Kleidukos.

This also needs Hackage revisions.

mpilgrem:
@sol, now that GHC 9.10.1 is released, may I release a version of `hpack` on Hackage that reflects this change?

sol:
> @sol, now that GHC 9.10.1 is released, may I release a version of `hpack` on Hackage that reflects this change?

@mpilgrem yes, please!

Releases are automated (https://github.com/sol/hpack/blob/main/.github/workflows/publish.yml). The only thing you have to do is bump  the package version + update the change log.

PR #576, opened 2024-04-10
Fix #575 Accommodate `GHC2024`

### LINKED ISSUE #575: Accommodate GHC2024
With https://discourse.haskell.org/t/ghc-9-10-1-alpha2-is-now-available/9197 and `GHC2024` on the horizon, Hpack should accommodate `GHC2024`.

mpilgrem:
This only requires an update to `README.md` because, in fact, Hpack itself already accepts any string for the `language`.


sol (approved):
Thanks @mpilgrem 🙏

PR #587, opened 2024-06-22
Fix #355 Add support for `includes`

### LINKED ISSUE #355: Why does hpack not use the `includes` property?
Cabal has the `includes` property: https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-includes

Which allows you to specify the C headers that should be included when we are compiling C code. How come this is ignored by Hpack? How does Hpack specify that C headers are needed when compiling some C code with `c-sources`?

mpilgrem:
Follows the pattern of existing `install-includes`. Fixes:
* #355

EDIT: A test added, again following `install-includes`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changes in 0.37.0
   - Add support for `asm-options` and `asm-sources` (see #573)
+  - Add support for `includes` (see #355)
 
 ## Changes in 0.36.1
   - Allow `Cabal-3.12.*`
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ values are merged with per section values.
 | `extra-lib-dirs` | · | | |
 | `extra-libraries` | · | | |
 | `include-dirs` | · | | |
+| `includes` | · | | |
 | `install-includes` | · | | |
 | `frameworks` | · | | |
 | `extra-frameworks-dirs` | · | | |
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -192,7 +192,7 @@ packageDependencies Package{..} = nub . sortBy (comparing (lexicographically . f
     deps xs = [(name, info) | (name, info) <- (Map.toList . unDependencies . sectionDependencies) xs]
 
 section :: a -> Section a
-section a = Section a [] mempty [] [] [] Nothing [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
+section a = Section a [] mempty [] [] [] Nothing [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] [] Nothing [] mempty mempty []
 
 packageConfig :: FilePath
 packageConfig = "package.yaml"
@@ -296,6 +296,7 @@ data CommonOptions asmSources cSources cxxSources jsSources a = CommonOptions {
 , commonOptionsExtraFrameworksDirs :: Maybe (List FilePath)
 , commonOptionsFrameworks :: Maybe (List String)
 , commonOptionsIncludeDirs :: Maybe (List FilePath)
+, commonOptionsIncludes :: Maybe (List FilePath)
 , commonOptionsInstallIncludes :: Maybe (List FilePath)
 , commonOptionsLdOptions :: Maybe (List LdOption)
 , commonOptionsBuildable :: Last Bool
@@ -333,6 +334,7 @@ instance (Semigroup asmSources, Semigroup cSources, Semigroup cxxSources, Semigr
   , commonOptionsExtraFrameworksDirs = Nothing
   , commonOptionsFrameworks = Nothing
   , commonOptionsIncludeDirs = Nothing
+  , commonOptionsIncludes = Nothing
   , commonOptionsInstallIncludes = Nothing
   , commonOptionsLdOptions = Nothing
   , commonOptionsBuildable = mempty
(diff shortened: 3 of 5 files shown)


sol (commented):
This looks good, just one question:

https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-includes says that this can be used for *"both header files that are already installed on the system and to those coming with the package to be installed"*.

@mpilgrem  for includes that come with the package, did you try if `cabal` includes those into the tarbal?  Or does `cabal` require you to also list them under `extra-source-files`?

If it's the latter, then we would need to address this somehow to be consistent with [don't repeat yourself](https://github.com/sol/hpack?tab=readme-ov-file#design-principles).

mpilgrem:
@sol, as far as I can tell:

* for Cabal's `install-includes: <filename list>`, `sdist` will include the files in the archive and Cabal will copy the files to `$libdir/includes` when the package is installed; and
* for Cabal's `includes: <filename list>`, `sdist` will not include the files in the archive. If you want one or more of the files to be the archive, you need to list them under Cabal's `extra-source-files:`.

So, perhaps Hpack does need two keys:

* one to correspond to Cabal's `includes:` only (perhaps named `includes-system`); and
* the other to correspond to Cabal's `includes:` and Cabal's `extra-source-files:` used together (perhaps named `includes-package`),

and, presumably, some form of de-duplication so that a file specified under Hpack's `includes-package` (if that key name is adopted) and Hpack's `extra-source-files` does not get rendered twice under Cabal's `extra-source-files`.

mpilgrem:
Rebasing on `main` branch.

sol:
> > The former files [header files that are already installed on the system] should be found in absolute paths, while the latter files [header files coming with the package to be installed] should be found in paths relative to the top of the source tree or relative to one of the directories listed in include-dirs.
> 
> If 'should be' is to be read as 'must', then perhaps Hpack could assume that all absolute file paths are system headers and all relative paths are to be included in `extra-source-files`.

Yes, that's what I was hoping for. However, giving it more thought, I'm not sure if this is a sound assumption.  My naïve assumption would be that a relative path actually can refer to a system include depending on what `-I` options are passed to the C compiler, or even if a corresponding include file is on any system standard directory for include files.

I still imagine that an algorithm could be, for each relative file name listed under `include`

1. If the file exists relative to `package.yaml` then add it to `extra-source-files`
1. For each relative directory listed under `include-dirs` If the file exists relative to the directory then add it to `extra-source-files`

If there are multiple hits, we add all of them, and leave resolving of precedence to `cabal`/`gcc`.  I think this ensures that an in-place build and a `cabal install` use the same includes.

One downside of this approach is that it's complicated to implement, requiring similar code as what we have for module inference.

sol:
Taking a step back, how crucial is this feature?

The Cabal documentation says that they are used for `compilations via C`.  Traditionally, in the context of GHC, "via-C" means unregistered builds / bootstrapping, which is pretty exotic, so I kind of doubt that this is what the Cabal docs refer to.

So what I kind of assume this is actually doing is pass `-include` to the C compiler when compiling c-sources.  This is still pretty exotic though, as usually you would just use `#include`-statements instead.

Where `include` could actually be somewhat useful is, if Cabal would also pass them to `ghc` via `-optP-include`.  Not sure if that's the case, the docs don't say anything about it and I'm currently away from the computers, so I can't try.

sol:
> @sol, I agree that it is complicated to implement.

Actually, I think I was wrong on the "complicated".  I think you can just take the total sum of all `includes` and `include-dirs` (from all sections and conditional branches) and the proceed as I described above.  This may, as a result include more `.h` files then necessary (as some combinations of `includes` and `include-dirs` may never occur together), but again it shouldn't be harmful.

> I am happy to give it a go, but it could take me some time.

Great 😊 

> Could you tolerate a step-by-step approach?

`main` should always be in a releasable state.  So I think better do it as one PR, but again, I don't think it's as complicated as I originally thought.

mpilgrem:
@sol, I had been assuming - perhaps incorrectly - that it was needed whenever a Haskell package used C code. My 'template' had been the Cabal file for `Win32`, eg https://hackage.haskell.org/package/Win32-2.14.0.0/Win32.cabal. Perhaps there is something 'special' about the `Win32` package.

sol:
> @sol, I had been assuming - perhaps incorrectly - that it was needed whenever a Haskell package used C code.

Definitely not needed, see  e.g. https://github.com/sol/simdutf-haskell/blob/main/package.yaml#L15

> My 'template' had been the Cabal file for `Win32`, eg https://hackage.haskell.org/package/Win32-2.14.0.0/Win32.cabal. Perhaps there is something 'special' about the `Win32` package.

I am gonna look at it

mpilgrem:
Here is another example of a package using `includes:`: https://hackage.haskell.org/package/mintty-0.1.4/mintty.cabal

mpilgrem:
I have seen something that implied to me that `includes:` may be required when header files are not co-located with `c-sources:` etc.

sol:
> I have seen something that implied to me that `includes:` may be required when header files are not co-located with `c-sources:` etc.

I don't think so, that is what `include-dirs` is for.

sol:
Looking at https://github.com/ghc/packages-Win32/commit/2a56e451262473de3c5efbbd3725633650e57471#diff-851be7da3fa86b5b2d90284d7440f3d15f9749698c2c14d02013641124050943R100

It's not a apparent  to me what the benefit of `includes` could possibly be.

sol:
> Here is another example of a package using `includes:`: https://hackage.haskell.org/package/mintty-0.1.4/mintty.cabal

Again, it's not a parent to me what `includes` would achieve here, given: https://github.com/RyanGlScott/mintty/blob/master/src/System/Console/MinTTY/Win32.hsc#L57-L59.

@RyanGlScott do you know for reason `includes` would be necessary here?

mpilgrem:
The original Cabal specification includes `includes:`, with the note `not yet used`: `https://downloads.haskell.org/cabal/Cabal-1.0/doc/pkg-spec.pdf. So, still tracking down when it was first documented.

sol:
From what I have tried, `includes` is neither passed to `ghc` (via `-optP-include`) nor to `gcc` (via `-include`).

I guess somebody has to go to the Cabal source and dig up the truth 😅

RyanGlScott:
> do you know for reason `includes` would be necessary here?

I'm not sure I know enough to say for sure. Ultimately, the code in `mintty` was adapted from `win32`, which was already using `includes` before `mintty` came along.

mpilgrem:
@sol, having worked through the history above, the `includes:` field of Cabal files have been ignored by GHC since GHC 6.10.1 (November 2008). So, I amn going to:
* raise a pull request to update the Cabal User Guide; and
* close this pull request.

PR #603, opened 2025-02-23
Implement `extra-files`

### LINKED ISSUE #602: Implement `extra-files`
`cabal-version: 3.14` of the Cabal package description format specification has [introduced](https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-3-14) new field [`extra-files`](https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-extra-files) - like `extra-source-files` but without the semantics that the files listed are tracked by build tools.

I'll raise an associated pull request.

mpilgrem:
`cabal-version: 3.14` of the Cabal package description format specification has [introduced](https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-3-14) new field [`extra-files`](https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-extra-files) - like `extra-source-files` but without the semantics that the files listed are tracked by build tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ library:
 | `build-type` | · | `Simple`, or `Custom` if `custom-setup` exists | Must be `Simple`, `Configure`, `Make`, or `Custom` | | |
 | `extra-source-files` | · | | Accepts [glob patterns](#file-globbing) | | |
 | `extra-doc-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.21.2` |
+| `extra-files` | · | | Accepts [glob patterns](#file-globbing) | | UNRELEASED |
 | `data-files` | · | | Accepts [glob patterns](#file-globbing) | | |
 | `data-dir` | · | | | | |
 | `github` | `source-repository head` | | Accepts `owner/repo` or `owner/repo/subdir` | `github: foo/bar` |
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -160,6 +160,7 @@ package name version = Package {
   , packageFlags = []
   , packageExtraSourceFiles = []
   , packageExtraDocFiles = []
+  , packageExtraFiles = []
   , packageDataFiles = []
   , packageDataDir = Nothing
   , packageSourceRepository = Nothing
@@ -602,6 +603,7 @@ data PackageConfig_ library executable = PackageConfig {
 , packageConfigFlags :: Maybe (Map String FlagSection)
 , packageConfigExtraSourceFiles :: Maybe (List FilePath)
 , packageConfigExtraDocFiles :: Maybe (List FilePath)
+, packageConfigExtraFiles :: Maybe (List FilePath)
 , packageConfigDataFiles :: Maybe (List FilePath)
 , packageConfigDataDir :: Maybe FilePath
 , packageConfigGithub :: Maybe GitHub
@@ -831,6 +833,7 @@ ensureRequiredCabalVersion inferredLicense pkg@Package{..} = pkg {
         makeVersion [2,2] <$ guard mustSPDX
       , makeVersion [1,24] <$ packageCustomSetup
       , makeVersion [1,18] <$ guard (not (null packageExtraDocFiles))
+      , makeVersion [3,14] <$ guard (not (null packageExtraFiles))
       , packageLibrary >>= libraryCabalVersion
       , internalLibsCabalVersion packageInternalLibraries
       , executablesCabalVersion packageExecutables
@@ -1024,6 +1027,7 @@ data Package = Package {
 , packageFlags :: [Flag]
 , packageExtraSourceFiles :: [Path]
 , packageExtraDocFiles :: [Path]
+, packageExtraFiles :: [Path]
 , packageDataFiles :: [Path]
 , packageDataDir :: Maybe FilePath
 , packageSourceRepository :: Maybe SourceRepository
@@ -1272,6 +1276,7 @@ toPackage_ dir (Product g PackageConfig{..}) = do
 
   extraSourceFiles <- expandGlobs "extra-source-files" dir (fromMaybeList packageConfigExtraSourceFiles)
   extraDocFiles <- expandGlobs "extra-doc-files" dir (fromMaybeList packageConfigExtraDocFiles)
+  extraFiles <- expandGlobs "extra-files" dir (fromMaybeList packageConfigExtraFiles)
 
   let dataBaseDir = maybe dir (dir </>) packageConfigDataDir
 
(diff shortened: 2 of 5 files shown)


test/EndToEndSpec.hs
```diff
@@ -638,6 +638,39 @@ spec = around_ (inTempDirectoryNamed "my-package") $ do
...
+            CHANGES.markdown
+            README.markdown
+        |]) {packageCabalVersion = "3.14"}
+
+      it "warns if a glob pattern does not match anything" $ do
+        [i|
+        name: foo
+        extra-files:
+          - "*.markdown"
+        |] `shouldWarn` ["Specified pattern \"*.markdown\" for extra-files does not match any files"]
```
sol, line 672: We only need one basic test case here.

Rational:

When you do TDD, you don't test everything exhaustively.  You only write as many tests as you need to drive the implementation. In this specific case we reuse a lot of existing code, which is already covered by other tests.  

Technically we  could add 10, or a 100, or a 1000 `extra-whatever-files`, reusing the existing code.  That does not mean that we would need to add x3 test cases for those.

One test case is enough here.

```haskell
describe "extra-files" $ do
  it "accepts extra-files" $ do
    ...
```

sol: Now, there is this potential risk that `extra-doc-files` could be removed in the future, and then we lose test coverage for `extra-files`.

If you wanna guard against this, then what you want to do is make those three test cases reusable and then reuse the same test code to test those fields.

If that's the approach you wanna take, then please feel free to give it a try.

README.md
```diff
@@ -143,6 +143,7 @@ library:
 | `build-type` | · | `Simple`, or `Custom` if `custom-setup` exists | Must be `Simple`, `Configure`, `Make`, or `Custom` | | |
 | `extra-source-files` | · | | Accepts [glob patterns](#file-globbing) | | |
 | `extra-doc-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.21.2` |
+| `extra-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.39.0` |
```
sol, line 146: As this is a very non-invasive change, do you think we should release as `0.38.1`?

NB: Adding a new field to `Hpack.Config.Package` I don't consider a "breaking" change:
- `Hpack.Config` is semi-internal
- A user should use `Hpack.Config.package` to construct a `Package` anyway, and if they did that, their code will continue to work just fine.
- In general, for records with a large number of fields, I as a user should always be prepared that new fields are added.

sol: Please also bump the package version accordingly, so that a release to Hackage is triggered

sol (changes requested):
Hey, thanks for looking into this.

Personally I am not convinced that `extra-files` offers enough value over `extra-source-files` and `extra-doc-files` to be worthwhile.

That said, this is a minimal change, so in this specific case I am willing to accept this patch for parity with Cabal.

Note that parity with Cabal is not a [stated design principle](https://github.com/sol/hpack?tab=readme-ov-file#design-principles) of Hpack, so this will always be a case by case decision.

sol:
@mpilgrem I updated the issue description, can you please update the commit message accordingly?

src/Hpack/Config.hs
```diff
@@ -828,7 +830,8 @@ ensureRequiredCabalVersion inferredLicense pkg@Package{..} = pkg {
 
     inferredVersion :: CabalVersion
     inferredVersion = fromMaybe packageCabalVersion $ maximum [
-        makeVersion [2,2] <$ guard mustSPDX
+        makeVersion [3,14] <$ guard (not (null packageExtraFiles))
```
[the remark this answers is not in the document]
mpilgrem: Will do.

test/EndToEndSpec.hs
mpilgrem: Happy to cut it down to one test. I can't see `extra-doc-files` being removed from package descriptions any time soon (there is no overlap between `extra-doc-files` and `extra-files`).

sol:
> Cabal-based build tools will treat `extra-files` differently (changes not tracked, ...) from `extra-source-files`

My question is more along the lines of "When does this actually matter?". 

I think in the worst case you  just rebuild too often.  Now if there is some extra file that changes all the time, then I would see a potential issue that `extra-files` could address.  It's just that I can't imagine a use case where this would matter.

README.md
sol, line 146:
```suggestion
| `extra-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.38.1` |
```

mpilgrem:
Rebuilding unnecessarily was given as the motivation for `extra-files`:
* https://github.com/agda/agda/issues/6502
* https://github.com/haskell/cabal/issues/8756
* https://github.com/haskell/cabal/issues/8817

but the fact it was not implemented for the first 20 years of Cabal suggests its need was not pressing!

test/EndToEndSpec.hs
```diff
@@ -638,6 +638,20 @@ spec = around_ (inTempDirectoryNamed "my-package") $ do
           - "*.markdown"
         |] `shouldWarn` ["Specified pattern \"*.markdown\" for extra-doc-files does not match any files"]
 
+    describe "extra-files" $ do
+      it "accepts a list of files" $ do
```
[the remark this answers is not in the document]
mpilgrem: We currently have:
~~~text
extra-doc-files
  accepts a list of files
~~~

and
~~~text
extra-files
  accepts extra-files
~~~

seemed a little 'circular' as an output - but if you are sure that is what you want, I'll put it through.

README.md
mpilgrem: The reason I've held off on the Hpack version is that there may be other changes soon that could be released together (and `extra-files` on its own is not a burning platform) - I'm currently (off repository) looking at moving #518 forward.

test/EndToEndSpec.hs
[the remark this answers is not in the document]
sol: The priorities are:

1. It needs to be clear what the purpose of a specific test case is ("What are we testing here?").
2. Consistency
3. Anything else

As (1) is the main priority, I would shy away from sacrificing it  for stylistic reasons.

In this specific case, we are testing that it "accepts extra-files".  We are not concerned with whether it can be a list, whether it accept globs, etc. (those aspects are covered by other test cases).

> seemed a little 'circular' as an output

Yes, definitely agree!

However, if you would want to address this, then I think the right approach would be to get rid of the `describe "extra-files"`.

But then again we already have things like:

```yaml
describe "spec-version" $ do
  it "accepts spec-version" $ do
```

So keeping the `describe` is at least consistent with what we already have (in line with our second priority).

README.md
sol: I say "release early, release often" and "minimize dependencies between tasks".

mpilgrem: Noted, but in this case #518 is on the heals of this one - and maybe #605 too, if I can convince you of its utility.

test/EndToEndSpec.hs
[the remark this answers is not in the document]
sol: @mpilgrem did you read this?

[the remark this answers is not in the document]
mpilgrem: Sorry, my mistake. I'll fix it now.

PR #605, opened 2025-03-11
Fix #604 When applicable, use `elif` in Cabal files

### LINKED ISSUE #604: When applicable, use `elif` in Cabal files
`elif` was introduced in `cabal-version: 2.2`. For example:

~~~yaml
when:
- condition: os(windows)
  then:
    source-dirs: windows
  else:
    when:
    - condition: "os(darwin) || os(linux)"
      then:
        source-dirs: unix-like
      else:
        source-dirs: unsupported-os
~~~

becomes:

~~~cabal
if os(windows)
  hs-source-dirs:
      windows
elif os(darwin) || os(linux)
  hs-source-dirs:
      unix-like
else
  hs-source-dirs:
      unsupported-os
~~~

I have a pull request that implements this.

mpilgrem:
See:
* #604


src/Hpack/Config.hs
```diff
@@ -880,6 +885,7 @@ ensureRequiredCabalVersion inferredLicense pkg@Package{..} = pkg {
               uses "RebindableSyntax"
           && (uses "OverloadedStrings" || uses "OverloadedLists")
           && pathsModule `elem` getMentionedModules sect)
+      , makeVersion [2,2] <$ guard (sectionSatisfies (isJust . maybeSectionAConditional) sect)
```
sol, line 888: What this is saying is that if a nested conditional can be transformed into an `elif`, then require Cabal 2.2.

What I think you would want to do is the other way around: If the rendered `.cabal` file has `cabal-version: 2.2` (or later) then produce `elif`s.

sol (changes requested):
This whole `HasEmpty` construction is not particularly beautiful.

I think what I would try is either
- try to transform nested conditionals into `elif`s at parsing time

or instead
- focus on some general solution to avoid staircases in `package.yaml` (as we started discussing in #604)

(or maybe even do both)

PR #613, opened 2025-04-01
Preserve empty lines in `description` when `cabal-version >= 3`

### LINKED ISSUE #612: Generate clean cabal descriptions
Since https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-3-0 , "Free text fields (e.g. description) preserve empty lines and indentation. In other words, you don’t need to add dots for blank lines."

Since 3.0 has been around for a long time, I'd like to start generating my cabal files with this cabal-version, and avoid adding the silly dots for blank lines, if possible. What's a good way to force hpack to select cabal-version 3.0 ?

sol:
(close #612)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## Unreleased
+## Changes in 0.38.1
   - Add support for `extra-files` (see #603)
+  - Preserve empty lines in `description` when `cabal-version >= 3` (see #612)
 
 ## Changes in 0.38.0
   - Generate `build-tool-depends` instead of `build-tools` starting with
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ library:
 | `build-type` | · | `Simple`, or `Custom` if `custom-setup` exists | Must be `Simple`, `Configure`, `Make`, or `Custom` | | |
 | `extra-source-files` | · | | Accepts [glob patterns](#file-globbing) | | |
 | `extra-doc-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.21.2` |
-| `extra-files` | · | | Accepts [glob patterns](#file-globbing) | | UNRELEASED |
+| `extra-files` | · | | Accepts [glob patterns](#file-globbing) | | `0.38.1` |
 | `data-files` | · | | Accepts [glob patterns](#file-globbing) | | |
 | `data-dir` | · | | | | |
 | `github` | `source-repository head` | | Accepts `owner/repo` or `owner/repo/subdir` | `github: foo/bar` |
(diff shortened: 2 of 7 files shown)


sol:
@simonmichael please try this and make sure that it works as expected.

simonmichael:
@sol, I confirm that if I use
```
verbatim:
  cabal-version: 3.0
```
hpack with this PR now generates description without the dots.
(I would be happy it also added a blank line after description's value, to distinguish it from the following fields, since it's usually multiple paragraphs.)

I'm not sure if using verbatim actually forces hpack to use the specified cabal-version. Eg could it still be using version 2.2 syntax elsewhere in the file ?

Even so yes I'd use it to get this more readable output. But I'd feel safer using some syntax in package.yaml that actually triggers selection of version 3.0 (but I didn't find a harmless way to do that; eg adding redundant commas in lists didn't force v3.0).

Perhaps ideally hpack would let us write an explicit cabal-version: field (without needing verbatim). And would warn if it detects syntax not compatible with that.

sol:
> (I would be happy if it also added a blank line after description's value, to distinguish it from the following fields, since it's usually multiple paragraphs.)

- This is only a cosmetic change, right? Or does this have any other relevance for you?
- Hpack currently lacks the machinery to allow for this.
- Playing devils advocate, would this even be strictly better in every situation for every user?

> I'm not sure if using verbatim actually forces hpack to use the specified cabal-version. Eg could it still be using version 2.2 syntax elsewhere in the file ?

No, it will only use syntax that is compatible with the specified `cabal-version`.  While `verbatim` fields are mostly handled transparently, Hpack sniffs for any `cabal-version` overrides to ensure that generated `.cabal` files are compatible.

> Even so yes I will use verbatim to get this more readable output, since other tools will probably warn me of issues as you say. But I'd feel safer using some syntax in package.yaml that actually triggers selection of version 3.0 (but I didn't find a harmless way to do that; eg adding redundant commas in lists didn't force v3.0).

You could technically specify `visibility: public` for your library component. This however will result in a `cabal check` warning + I can't guarantee you that Hpack will keep this behavior indefinitely.

> Perhaps ideally hpack would let us write an explicit cabal-version: field (without needing verbatim). And would warn if it detects syntax not compatible with that.

For Hpack `.cabal` files are a backend format.  Human readability is important, but being able to control every possible formatting related aspect is not an explicit design goal of Hpack.

While in general I'm happy to improve formatting, I'm not eager to add fields to `package.yaml` with the  sole motivation of influencing the formatting of generated `.cabal` files.

simonmichael:
Thanks!

PR #616, opened 2025-07-04
Add 'Hooks' to the build-types (since cabal 3.14)

lxsameer:
Cabal `>=3.14`, introduced `Hooks` build type that supposed to be the replacement for old `Setup.hs` stuff. This little patch just adds that to the `BuildType` data type so users can start using it.


mpilgrem:
@lxsameer, see `Hpack.Config.makeCabalVersion`.

PR #620, opened 2025-08-20
Fix #618 Infer cabal-version: 3.12 needed from PackageInfo_* modules

### LINKED ISSUE #618: Autogenerated module PackageInfo_* requires cabal version 3.12
My stack project fails to build because it can't find the module _PackageInfo_XXX_. `stack build` gives a warning message:

~~~
Error: [Cabal-5559] 
[autogen-guard] To use the autogenerated module PackageInfo_* you need to specify `cabal-version: 3.12` or higher.
~~~

Looks like the problem is that hpack sets `cabal-version: 2.0` even though the cabal-field `autogen-modules` is generated. Here's a simple hpack example:

~~~yaml
---

spec-version: 0.36.0

name: myproj
version: 0.0
language: GHC2024 
dependencies:
    - base
    - relude

default-extensions:
    - NoImplicitPrelude
    - OverloadedStrings

executables:
    myproj:
        main:        Main.hs
        generated-other-modules: 
            - Paths_myproj
            - PackageInfo_myproj

~~~

mpilgrem:
See:
* #618 
* https://github.com/haskell/cabal/pull/11171

This pull request does not (not yet?) add a unit test.


mpilgrem:
@sol, thanks for the review. I've cleared the points and added the tests.

Issue #638, opened 2026-01-17
cabal-version >= 3.4 makes internal-libraries fail


karamellpelle:
I added `verbatim: cabal-version: 3.12` to my _package.yaml_ in order to get [access to PackageInfo_XXX](https://github.com/sol/hpack/issues/618). However, this turned out to cause problems when specifying `internal-libraries ` (i.e. _cabal_ sublibraries) in _hpack_: [After cabal version 3.4 you need to specify sublibraries explicitily](https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-3-4).

This is fixed in my _package.yaml_ by  specifying dependencies to my internal libraries as  `mypackage:myinternal`. I first thought of this as a bug, but cabal says the purpose is to let users have a sublibrary with the same name as some external dependency.  So I'm not sure how _hpack_ will consider names of internal libraries? As now, _hpack_ should make it clear in the documentation that  references to `internal-libraries` fields must be written in the form `<mypackage>:<myinternal>` .

-----

References: https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#library .

PR #639, opened 2026-01-18
Re #638 Add docs on deps on main libs and sublibs, post-Cabal 3.4

mpilgrem:
See:
* #638

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,10 @@ specify the individual dependencies:
 Individual dependencies as objects are only supported from version
 `0.31.0`.
 
+When the Cabal file version is 3.4 or greater, a dependency on a named library
+component of the same package must use the `package-name:library-name` format
+introduced in Cabal file version 3.0.
+
 When a dependency is specified as an object, you can use the `mixin`
 field to control what modules from the dependency your program will
 see and how its signatures are filled in:
--- a/README.md
+++ b/README.md
@@ -406,6 +406,19 @@ specify the individual dependencies:
 Individual dependencies as objects are only supported from version
 `0.31.0`.
 
+Dependencies can be the main (unnamed) library of a package or named
+sublibraries of a package. For example, for a package named `my-package` with
+a main library and a sublibrary named `my-library`:
+
+* `my-package` or `my-package:my-package` specifies a dependency on its main
+  library only;
+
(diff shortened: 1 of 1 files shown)


sol:
Hmm, this is not really in line with how Hpack currently works.  A user of Hpack shouldn't usually be concerned about `cabal-version`.

Consider this:
- a user has a package that uses internal libraries
- at a later point the user adds `extra-files` to the package and as a result the package won't compile anymore

That's not obvious and feels wrong.  Not sure how we should address this.

---
Idk, cabal feels like an over-engineered Frankenstein, and nobody is closing the flood gates.

mpilgrem:
@sol, two possible dimensions to #638 are:
* 'documentation' of the current state of affairs (what this pull request seeks to address, only); and
* 'Hpack functionality'.

The underlying problem seems to me to be that when named libraries were introduced in Cabal 2.0, insufficient thought was given to possible future ramifications, and that was only sorted out in Cabal 3.4. It has also taken a while for Cabal's own documentation to catch up with its functionality.

I am conscious that the Hpack reference documentation seeks to not duplicate Cabal's own documentation. However, as a Hpack user had been caught by this, and it could be fixed with one sentence, it seemed to me that the additional of a sentence was tolerable.

I did think about 'Hpack functionality' but concluded that, given the Cabal 2.0 to 3.2 interlude, there was no way (or, at least, no pleasant way) for Hpack to know if a user really wanted to specify a dependency on a named library of the same package or the main library of another package. However, now that Cabal 3.4 is, itself, 'history' (as we now have Cabal 3.16), one approach might be:

"If a Hpack user specifies `internal-libraries`, then force `cabal-version` >= 3.4 and document that `package name:library name` is required for dependencies on named libraries."

In that case, there would no longer be any ambiguity from a Hpack perspective - but it would be a break from the past. Users expecting the Cabal 2.0 to 3.2 behaviour would encounter messages like:
~~~text
> stack build
testNamedLibs> configure (lib + sub-lib + exe)
Configuring testNamedLibs-0.1.0.0...
Error: [Cabal-8010]
Encountered missing or private dependencies:
    my-named-library
~~~

mpilgrem:
@sol, I have changed my approach to the documentation, in the light of the merger of:
* #651

PR #640, opened 2026-03-05
Use cryptohash-sha256, not crypton, for SHA256 hashes

mpilgrem:
My general motivation is to reduce indirect dependencies on `memory` and `basement` (which are unmaintained) (a goal I have for Stack). This eliminates a dependency on `crypton`.

However, in addition, `hpack` only uses `crypton` for SHA256 hashes. `cryptohash-sha256` is a lightweight package for that single purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## Changes in 0.39.2
+  - Depend on `cryptohash-sha256`, rather than `crypton`, for SHA256 hashes
+
 ## Changes in 0.39.1
   - Add support for `mhs-options` (MicroHs)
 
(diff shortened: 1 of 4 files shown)


sol:
@mpilgrem apparently released versions Hpack stopped building https://github.com/sol/hpack/actions/runs/22788497963/job/66110378136, which I think is unfortunate.

cc @kazu-yamamoto

sol:
`main` is also broken

mpilgrem:
@sol, I will look into that.

kazu-yamamoto:
One thing I found is that boundary of `tls` v2.2 is loose. It must use `hpke` v0.0.0, not v0.1.0. I will update the metadata of Hackage.

kazu-yamamoto:
@sol Sorry for your inconvenience. This is my fault. I have updated metadata of `tls` v2.2.2. Would you run CI again?

Issue #641, opened 2026-03-07
Hpack doesn't build (due to recent `memory` -> `ram` changes?)


sol:
@kazu-yamamoto @mpilgrem first things first, thank you for your effort towards migrating things away from unmaintained packages.

I understand that sometimes things don't go 100% smooth, for Hpack specifically, it fails to build, presumably due to these recent changes.  I don't have enough insight in what exactly is going wrong.

Can you look into this?

mpilgrem:
@sol, I am looking into this. In the case of `hpack-0.39.2`, my immediate solution is to put an upper bound on the `crypton` dependency.

PR #642, opened 2026-03-07
Re #640 Add an upper bound on crypton

mpilgrem:
See:
* #640
* #641
* https://github.com/snoyberg/http-client/pull/573

The immediate problem is that direct dependency `http-client-tls <= 0.3.6.4` does not support newly-released `crypton-1.1.0` (an indirect dependency of `hpack`) but does not, itself, specify an upper bound on `crypton`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Changes in 0.39.3
+  - Add upper bound `crypton < 1.1`, as dependency `http-client-tls <= 0.3.6.4`
+    does not support `crypton-1.1.0` but does not itself have an upper bound.
+
 ## Changes in 0.39.2
   - Depend on `cryptohash-sha256`, rather than `crypton`, for SHA256 hashes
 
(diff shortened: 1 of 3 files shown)


sol (approved):
As this fixes the build this is fine with me.

sol:
I guess we also need corresponding Hackage revisions.  I think you already have the appropriate permissions.  Do you wanna take care of this as well?

mpilgrem:
@sol, for the `hpack-0.39.*` packages on Hackage with a `crypton` dependency, I have edited the Cabal file to specify `crypton < 1.1`. `hpack-0.39.2` did not have a direct dependency on `crypton`, so a Cabal file revision is not possible.

mpilgrem:
@sol, the CI is green but the `publish` job fails (I think because `indexed-traversable-0.1.4` does not yet recognise that it supports GHC 9.14.1/`base-4.22.0.0`). Should I release manually?

Issue #483, opened 2022-03-09
Get ready for sublibraries.


philderbeast:
Please get ready for sublibraries. Cabal is mostly there. This is good for distribution of closely related libraries within a single package.

```cabal
name: pkgA

library
  build-depends: foo

library foo
  visible: True
```

SEE: https://github.com/haskell/cabal/issues/5660
SEE: https://github.com/brendanhay/amazonka/issues/671

tfausak:
I think this could work the same way that executables do. (See #161 for some commentary about that.) We could introduce a new top level `libraries` key that allows them to be named. Then we could make `library: ...` a shortcut for `libraries: { PACKAGE_NAME: ... }`.

philderbeast:
Should we rename these to `libraries`? With `visibility: public` they're not that internal anymore.

sol:
Yesterday, I pushed a change that allows to have aliases for fields: https://github.com/sol/hpack/commit/92c6c94115e71532239bf0dd249c97502b5d6641

Optionally, aliases can be deprecated.

This facilitates renaming in a backwards compatible way.

I would still want to see a working, fully self-contained example before deciding to do that.

Instead of renaming, we may also decide to add a new `libraries` field that sets `visibility: public` by default + deprecate `internal-libraries`. I think this would be more appropriate if public libraries are the common case, not sure.

As a side note, Cabal having both `exposed` / `visibility` is not a great user interface.

philderbeast:
Thanks for picking this up @sol. I went looking for `libraries` but somehow missed `internal-libraries`.

This alias work is amazing. I have an hpack **dependency**. Others have a cabal **build-dependency**. With this we might bridge the gap.

Could aliasing eventually cover all fields so that we might use hpack with a phrasing matching cabal's?

On 92c6c94, aliases seem second class when I see a test "gives the primary name precedence". Could we give this more of a "warns on duplicate fields" feel?

brandon-leapyear:
Another issue I found: you can explicitly depend on the main library by using the package name as a sublibrary name, which allows you to do:
```cabal
mylib:{mylib, internallib}
```

But hpack seems to strip it out:
```yaml
dependencies:
- mylib:{mylib, internallib}
```
```cabal
build-depends:
  mylib:{internallib}
```

https://fgaz.me/posts/2019-11-14-cabal-multiple-libraries/

mpilgrem:
[...]

PR #645, opened 2026-05-04
Re #483 Accept main lib in shorthand syntax for dependencies

mpilgrem:
See:
* https://github.com/sol/hpack/issues/483#issuecomment-1358722897

Currently, something like:
~~~yaml
spec-version: 0.36.0
name: myPackage
dependencies:
- base

library:
  source-dirs: src

internal-libraries:
  sublib1:
    source-dirs: sub1
  sublib2:
    source-dirs: sub2

executables:
  myExe:
    source-dirs: app
    main: Main.hs
    dependencies:
    - myPackage:{myPackage, sublib1, sublib2}
~~~

renders in the Cabal file as (extract):

~~~text
executable myExe
  main-is: Main.hs
  hs-source-dirs:
      app
  build-depends:
      base
    , myPackage:{sublib1,sublib2}   <<<< The reference to the main library component is missing
  default-language: Haskell2010
~~~

This pull request seeks to fix that. A new test is added.

### DIFF
--- a/src/Hpack/Syntax/Dependencies.hs
+++ b/src/Hpack/Syntax/Dependencies.hs
@@ -73,10 +73,15 @@ parseDependency :: Fail.MonadFail m => String -> Text -> m (String, DependencyVe
 parseDependency subject = fmap fromCabal . cabalParse subject . T.unpack
   where
     fromCabal :: D.Dependency -> (String, DependencyVersion)
-    fromCabal d = (toName (D.depPkgName d) (DependencySet.toList $ D.depLibraries d), DependencyVersion Nothing . versionConstraintFromCabal $ D.depVerRange d)
+    fromCabal (D.Dependency pkgName verRange libraries) =
+      (depName, DependencyVersion Nothing $ versionConstraintFromCabal verRange)
+      where
+        depName :: String
+        depName = prettyShow pkgName <> case DependencySet.toList libraries of
+          [D.LMainLibName] -> ""
+          [D.LSubLibName name] -> ":" <> prettyShow name
+          xs -> ":{" <> intercalate "," (map renderComponent xs) <> "}"
 
-    toName :: D.PackageName -> [D.LibraryName] -> String
-    toName package components = prettyShow package <> case components of
-      [D.LMainLibName] -> ""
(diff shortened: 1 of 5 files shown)


src/Hpack/Syntax/Dependencies.hs
```diff
@@ -79,4 +79,4 @@ parseDependency subject = fmap fromCabal . cabalParse subject . T.unpack
     toName package components = prettyShow package <> case components of
       [D.LMainLibName] -> ""
       [D.LSubLibName lib] -> ":" <> prettyShow lib
-      xs -> ":{" <> (intercalate "," $ map prettyShow [name | D.LSubLibName name <- xs]) <> "}"
+      xs -> ":{" <> (intercalate "," $ ( map prettyShow [ package | D.LMainLibName <- xs] <> map prettyShow [name | D.LSubLibName name <- xs])) <> "}"
```
sol: Untested:

```suggestion
      xs -> ":{" <> intercalate "," (map renderSubLibName xs) <> "}"
        where
          renderSubLibName :: D.LibraryName -> String
          renderSubLibName = \ case
            D.LMainLibName -> prettyShow package
            D.LSubLibName name -> prettyShow name
```

sol (changes requested):
This is a great fix, thanks for digging into this!

1. Semantically this is a correct fix.
1. In a way you could say the bug was that the existing code did not pattern match on `DependencyName`.  With that view, a proper fix will do pattern matching (see my suggestion below).  This way somebody who reads the code can immediately see that all cases are handled, and none of the constructors are silently ignored.  One additional benefit of this approach is that it keeps the order intact.
2. Technically we should probably also pattern match on `Dependency` in `fromCabal`.  https://hackage-content.haskell.org/package/Cabal-syntax-3.16.1.0/docs/Distribution-Types-Dependency.html#v:Dependency

src/Hpack/Syntax/Dependencies.hs
mpilgrem: I have adopted a version of this, in combination with pattern matching on `Dependency` in `fromCabal`.

sol (approved):

sol:
@mpilgrem please go ahead and merge when you're done.

PR #647, opened 2026-05-09
Add support for top-level field `codeberg`

sol:


sol:
@mpilgrem can you give this a quick review?

mpilgrem:
Looks good to me. I did wonder about the name of the type `GitHub` and the use of its single constructor `GitHub` to represent both GitHub-related and Codeberg-related values. However, I suppose one could say that Codeberg has copied a 'GitHub' structure.

sol:
Yes, they use the same URL layout, so we can reuse the existing code.

Issue #649, opened 2026-06-05
Assume at least `cabal-version: 3.4` when using `pkg:lib` syntax


mpilgrem:
The Cabal project's introduction of `pkg:lib` syntax for dependencies in `cabal-version: 3.0` involved a misstep that was fixed in `cabal-version: 3.4`. See:
* https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-3-4

I suggest that future versions of Hpack should assume that if a user makes use of `pkg:lib` syntax, they intend to use at least `cabal-version: 3.4` and not (as presently assumed) `cabal-version: 3.0`.

Not least, that is because `Cabal-3.4.0.0`, released 2021-02-06, is itself over 5 years old. That is, I doubt if anyone using Hpack today really wishes to pick `cabal-version: 3.0` over `cabal-version: 3.4`.

The motivation is that I find myself adding `verbatim:` because Hpack is picking the 'broken' version of the syntax.

sol:
> The motivation is that I find myself adding `verbatim:` because Hpack is picking the 'broken' version of the syntax.

How exactly is that? From what I understand, 3.4 syntax is more restrictive.  3.4 will only benefit you if you have a sublibrary and you want to depend on some external dependency with the exact same name as you sublibrary. .
My naïve assumption is that this almost never happens, but I may very well be wrong.

- Is this more common than I would naïvely assume?
- Is this what you are trying to do? Example?
- Did I miss anything else?

`verbatim` is really just meant as an emergency hatch, so If you can motivate this with a real word example then I'm probably on board with bumping this to `3.4`.

mpilgrem:
I only have toy package examples. I was playing with:
~~~yaml
spec-version: 0.36.0

name: myPackage

dependencies:
- base

internal-libraries:
  acme-missiles:
    source-dirs: sub
    dependencies:
    - acme-missiles
~~~

where the sublibrary `acme-missiles` adds something to the 'original' `acme-missiles` package on which it depends.

With `cabal-version: 3.0` this is interpreted as a cyclical dependency. With `cabal-version: 3.4` the dependencies `acme-missiles` and `myPackage:acme-missiles` are distinguished.

A 'real life' example of this (for a short while; the sublibrary was later renamed) was [`haskell-debugger`](https://github.com/well-typed/haskell-debugger) where a sublibrary `dap` depended on a package `dap`.

sol:
[...]

PR #650, opened 2026-06-06
Fix #648 Cabal version depends on custom setup deps syntax

### LINKED ISSUE #648: Wrong `cabal-version` with certain `custom-setup:` dependencies
This `package.yaml`:
~~~yaml
spec-version: 0.36.0

name: testHpack

custom-setup:
  dependencies:
  - base
  - myPackage:myPackage-sublib
~~~

generates `testHpack.cabal`:
~~~test
cabal-version: 1.24

-- This file has been generated from package.yaml by hpack version 0.39.5.
--
-- see: https://github.com/sol/hpack

name:           testHpack
version:        0.0.0
license:        BSD3
license-file:   LICENSE
build-type:     Custom

custom-setup
  setup-depends:
      base
    , myPackage:myPackage-sublib
~~~

and, so, causes Cabal (the library) to complain:
~~~text
unexpected Sublibrary dependency syntax used. To use this syntax the package needs to specify
at least 'cabal-version: 3.0'. Alternatively, if you are depending on an internal library, you
can write directly the library name as it were a package.
~~~

(Discovered because I am currently working on Stack accepting dependency packages with one or more public sublibraries and no main library, like Cabal (the tool) now does.)

mpilgrem:
See:
* #648

Also bumps the LTS snapshot in `stack.yaml`.

Not set up for immediate release if merged, because a #649 fix (trivial) may go hand-in-hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Changes in UNRELEASED
+  - Infer `cabal-version: 3.0` when custom setup has sublibraries as
+    dependencies.
+
 ## Changes in 0.39.6
   - Add support for top-level field `codeberg`
 
--- a/src/Hpack/Config.hs
+++ b/src/Hpack/Config.hs
@@ -836,6 +836,7 @@ ensureRequiredCabalVersion inferredLicense pkg@Package{..} = pkg {
     inferredVersion = fromMaybe packageCabalVersion $ maximum [
         makeVersion [2,2] <$ guard mustSPDX
       , makeVersion [1,24] <$ packageCustomSetup
+      , subLibDepsSyntaxCabalVersion <$ (guard . hasAnySubcomponents . customSetupDependencies =<< packageCustomSetup)
       , makeVersion [1,18] <$ guard (not (null packageExtraDocFiles))
       , makeVersion [3,14] <$ guard (not (null packageExtraFiles))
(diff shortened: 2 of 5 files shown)


sol:
> Not set up for immediate release if merged, because a #649 fix (trivial) may go hand-in-hand.

1. In general I follow "release early, release often".  That means every change that is a clear improvement can be released immediately.
1. I think this change can go into a minor version, #649 however will require a major version bump.

sol (changes requested):
This looks great 👍🎉

Can you:
- add a test
- if you agree that this can go into a minor version, then bump the minor version

mpilgrem:
@sol, where should such a test be located? I had looked for similar tests (by seaching for `cabal-version`) but none jumped out at me.

sol:
> @sol, where should such a test be located? I had looked for similar tests (by seaching for `cabal-version`) but none jumped out at me.

Probably right after this test case:

https://github.com/sol/hpack/blob/5b9cddf27ba67f1c2c23c4c1befa6a64aaaa90e5/test/EndToEndSpec.hs#L1291

see also:
https://github.com/sol/hpack/blob/5b9cddf27ba67f1c2c23c4c1befa6a64aaaa90e5/test/EndToEndSpec.hs#L918

PR #651, opened 2026-06-07
Fix #649 cabal-version: 3.4 for sublibrary dependencies

### LINKED ISSUE #649: Assume at least `cabal-version: 3.4` when using `pkg:lib` syntax

mpilgrem:
See:
* #649

This assumes that the change in behaviour for sections requires a major version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## Changes in UNRELEASED
-  - Infer `cabal-version: 3.0` when custom setup has sublibraries as
-    dependencies.
+## Changes in 0.40.0
+  - Infer `cabal-version: 3.4` (rather than `cabal-version: 3.0`) when a section
+    has sublibraries as dependencies. Make the same inference when custom setup
+    has such dependencies.
 
 ## Changes in 0.39.6
   - Add support for top-level field `codeberg`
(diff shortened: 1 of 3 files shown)


sol (approved):
LGTM

Can you please bump the version.  Or do you plan to get more things into that release?

sol:
Anyway, I'm merging this now. Please open a new PR to bump the version.

sol:
@mpilgrem please hold off with a release.  We need to figure out what to do about #638 / #639 first.

PR #652, opened 2026-06-08
Refactor

sol:


test/EndToEndSpec.hs
```diff
@@ -1315,18 +1315,18 @@ spec = around_ (inTempDirectoryNamed "my-package") $ do
         custom-setup:
           dependencies:
             - pkg:lib
-        |] `shouldRenderTo` customSetupWithCabalVersion "3.4" [i|
+        |] `shouldRenderTo` (customSetup [i|
         setup-depends:
             pkg:lib
-        |]
+        |]) { packageCabalVersion = "3.4" }
```
sol, line 1321: @mpilgrem I do think it's better to stick to the existing style, if only for consistency.

mpilgrem: Sure. I had not appreciated that was the existing style.

